PR #26, opened 2016-08-30
Add Spanish translation

larjona:
Thank you for working in simple gallery. I am a happy user.
Here you are the Spanish translation.
Best regards


tibbi:
thank you :) Ive updated the file, items marked as translatable="false" must be present only at the default english file.

Issue #35, opened 2016-10-08
Wrong dirction of swap in Arabic and other RTL languages


xsoh:
In RTL languages devices (e.g. Arabic), when opining an Image or video the swap between them shall be from right to left. While English and other LTR are from left to right. 

The application doesn't swap to the right way in RTL language devices.

PR #36, opened 2016-10-08
FIX: Wrong direction of swap in Arabic and other RTL languages #35

xsoh:
This will do the job for both RTL and LTR


tibbi:
perfect, thanks

tibbi:
@xsoh is it okay to have RTL layout enabled with english language at all? Shouldn't RTL be available only if the whole app is translated to Arabic?

xsoh:
@tibbi Old Android devices (below API level 18) doesn't support RTL, only translations. 
Also mostly no one use RTL with English translation. I've tested the app with English Language (i.e. LTR) and it worked.

PR #51, opened 2016-11-10
Fixed #50 for reals

### LINKED ISSUE #50: Silent crash when deleting photos on an external SD card
When I try to delete some photos stored in my external SD card (more specifically gifs), the application silently crashes and self-closes. The app never asked me for SD card access (I'm using Android 5.0 Lollipop), and thus the photos remain untouched.

XyLoNaMiyX:
`treeUri` needs to be set at least once, always (despite we already have write access or not).

### DIFF
--- a/app/src/main/java/com/simplemobiletools/gallery/Utils.kt
+++ b/app/src/main/java/com/simplemobiletools/gallery/Utils.kt
@@ -171,7 +171,9 @@ class Utils {
         }
 
         fun isShowingWritePermissions(activity: Activity, file: File): Boolean {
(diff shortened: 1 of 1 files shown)


XyLoNaMiyX:
@tibbi I haven't tested this on pre-KitKat devices, so I don't know if there it would also ask for SD card access. Also, there's the bug from many popups coming when asking for deletion (which I presume was also before, because it happens when it asks for permission).

tibbi:
Kitkat its ok, Ive checked it. Ive removed the file.canWrite check, its apparently another function that just wont work properly. Retest when you can pls.

XyLoNaMiyX:
I'll try it again tomorrow!

tibbi:
alright thanks, good enough

PR #54, opened 2016-11-13
simplified  chinese translation

tcg2008:
add simplified chinese translation


tibbi:
Hello, I can accept and update it, but please try to keep the same order of strings as in the original file, do not sort them alphabetically. Thanks :)

PR #85, opened 2016-12-10
Update Russian

mcbaze:
new item translated.
playstore description translation variant added (please check it to be sure everything is right)


tibbi:
i guess its ok, thanks :)

PR #116, opened 2016-12-29
Update Chinese simplified translation

KrasnayaPloshchad:
This pull request will translate the description into simp. Chinese.


tibbi:
cool thanks

PR #119, opened 2017-01-01
Translating partly in french

RyDroid:


tibbi:
please remove the copyright and uncomment all strings, it cannot be compiled like this

RyDroid:
Ok, I did it.

PR #129, opened 2017-01-06
Update strings.xml

j-biernacki:
I took a look on the link you gave, I have some remarks about the translation :
The apps name is "Gallery" in English, the French word is literally "Galerie", but... I think the real name could be Album... I translated sometime Album in the text, but let "Galerie" in the name. I can literally translate a name, but I can not change a name. If you want to change the name, you can, but it is not my decision.
I have a small doubt on the third long description paragraph, but anyway, this is totally translated. Have fun !


tibbi:
nice, thanks! I think Galerie is better

PR #186, opened 2017-03-10
Option for max brightness when viewing media

nvllsvm:
Provides an option to temporarily set the screen brightness to maximum when viewing media.

This only covers the `ViewPagerActivity` though it might also make sense elsewhere in the app. Thoughts?


tibbi:
Hey, nice one! I will check the user experience later. I generally dont like when apps mess with my brightness or volume, but a defaultly disabled checkbox cant hurt much. A few people requested it already.

tibbi:
ye looks ok, thanks for the pull request!

PR #188, opened 2017-03-10
Update Spanish translation (complete all except the release notes)

larjona:
I've completed the Spanish translation (except the release notes).
I'll try to commit those in a different PR.


tibbi:
nice, thanks! You dont need to translate the release notes, they will not be shown anywhere anymore :) I might add some note to them. Ill take a look at these strings a bit later.

tibbi:
@larjona Ive made release notes non translatable, do not translate them please. They are shown only at the first launch after the app update, so translations are always late anyway.

PR #339, opened 2017-08-27
Fixes Crash on Android API 16 (Jelly Bean)

Gan0k:
Check that the version is greater than API 16 when calling
Activity.isDestroyed()


tibbi:
nice, thanks :)

PR #348, opened 2017-09-02
Updates for version 2.13.4

gregory678:
PS: Please update Google Play Store description to one from this file. Current one is looking like machine-made.


tibbi:
heh, Im not good at writing such texts. However Ive done some AB testing and found out that users dont read the app descriptions anyway, so it doesnt really matter whats there. Thats my case too, when I install an app I check the screenshots and ratings, dont read the descriptions. Maybe just the short one.

PR #356, opened 2017-09-09
update german translations

varakh:


tibbi:
Hey, I received some newly translated german strings some time ago and Ive updated them, please check which strings still need to be updated, if any. Maybe opening a new pull request would be the simplest solution. Thanks

varakh:
Done

tibbi:
cool, thanks

PR #398, opened 2017-10-20
Chinese translation for v2.16.0

Pzqqt:
Update chinese translation for v2.16.0 .
I love this app !


tibbi:
cool, thanks

tibbi:
Hey @Pzqqt , I had to revert this commit, as it messed up the string order, and added many extra strings. To translate what's needed please go to https://github.com/SimpleMobileTools/Simple-Gallery/blob/master/app/src/main/res/values-zh-rCN/strings.xml , press the pencil at the top right corner, then just edit the strings that are needed. Thanks

PR #413, opened 2017-10-26
Update strings.xml

smarquespt:


app/src/main/res/values-pt/strings.xml
```diff
@@ -127,8 +127,8 @@
     <string name="allow_video_gestures">Permitir controlo do volume e brilho dos vídeos através de gestos verticais</string>
     <string name="show_media_count">Show folder media count on the main view</string>
```
smarquespt: What is counted? Number of files from folder?

tibbi: it is a bit more specific, it counts the number of currently displayed media files in the folder. So if nothing is filtered out, its the sum of images, videos and gifs. If gifs are filtered out, its just the sum of images and videos. Its basically the amount of items that you should see once you click on the folder.

PR #416, opened 2017-10-28
Update strings.xml

Pzqqt:
Add missing strings translation & fixed some translations


tibbi:
Hey, all those missing strings are probably in the Commons library, as they are reused between multiple projects. Try translating them in this file https://github.com/SimpleMobileTools/Simple-Commons/blob/master/commons/src/main/res/values-zh-rCN/strings.xml . Thanks

tibbi:
guess I can close this, all strings were moved in Commons

PR #427, opened 2017-11-01
Lossless rotation for JPEG

ltGuillaume:
Tested on Android 7.1.1 VM and Android 4.4 VM


app/src/main/kotlin/com/simplemobiletools/gallery/activities/ViewPagerActivity.kt
```diff
@@ -483,18 +496,45 @@ class ViewPagerActivity : SimpleActivity(), ViewPager.OnPageChangeListener, View
...
+                6 -> 90f
+                else -> 0f
+            }
+            orientationDegrees = (orientationDegrees + mRotationDegrees) % 360
+            orientation = when (orientationDegrees) {
+                90f -> 6
+                180f -> 3
+                270f -> 8
+                else -> 1
+            }
```
tibbi: for these 2 "when" cases create some helper functions similar to https://github.com/SimpleMobileTools/Simple-Gallery/commit/3c3740731564b47b1be0e9c608b7852a2cc2eef2#diff-0c323285093b03e42d5770c0ff050227R260 , where instead of magic numbers 6,3,8,1 you use constants like ExifInterface.ORIENTATION_ROTATE_90 , degrees are ok. You can also cast orientation to string in the function, to avoid ugly code like: ""+ orientation

tibbi:
Ive added 2 comments. Havent checked the actual functionality yet, just the code style. Thanks

app/src/main/kotlin/com/simplemobiletools/gallery/activities/ViewPagerActivity.kt
ltGuillaume: - Haha I had that,but changed it to hhardcoded values for uniformity with https://github.com/SimpleMobileTools/Simple-Gallery/blob/3c3740731564b47b1be0e9c608b7852a2cc2eef2/app/src/main/kotlin/com/simplemobiletools/gallery/fragments/PhotoFragment.kt#L127

- I can make that function public to prevent redundancy or create a helper function, your call.
- Will also check if file is copied successfully beefore EXIF operation.
- Should I use an abstract copy function of yours maybe?

tibbi: just use android.media.ExifInterface for constants, they fit there.

tibbi:
yea, from style point of view it looks much better. However, EXIF seems to work properly only on internal storage, it is ignored at SD card files. So at isJpg() you should also check !isPathOnSD(currPath).
So basically we will be using EXIF rotation only at jpgs on internal storage. Do the jpg and sd card check earlier (together with a File(it).exists() check), so we can avoid the whole temporary file creation and inputstreams altogether, if we are just modifying the exif of an already existing image. Thanks

ltGuillaume:
How have you checked this? I want to reproduce, but I can't: on SD it's working here.

tibbi:
currently it doesnt even get to the exif part, it fails at saveRotation FileNotFoundException with file not found, permission denied. What Android version are you testing the SD card with? Do you have an actual device, or emulators only?

ltGuillaume:
If we put the temporary file in the app's own directory
`val tmpFile = File(applicationInfo.dataDir, ".tmp_${it.getFilenameFromPath()}")`
then do the changes and then move it to the selected path, it should work. The current
```
renameFile(tmpFile, selectedFile) {
  deleteFile(tmpFile) {}
}
```
won't work then. I don't know what kind of copyMoveFiles()/copyMoveTo() kind of abstract function would be best for that, though, I've been looking through it, but that means I'd have to figure out the entire structure of Gallery, Common-Tools etc., not to mention I'm not used to Kotlin.

tibbi:
yeah, guess I will accept the current version and finish it myself, working with SD card files on android is a nightmare, not really a good place to start :p Thanks for the work

ltGuillaume:
Yeah it is. But basically I don't want to "re-implement" stuff that's already there, that won't help you in any way. I just don't have a clear picture of the app's structure yet.

tibbi:
ye that's understandable, many tricks are needed at working with files

tibbi:
so I ended up rewriting it quite a bit as it had a few issues, seems to work fine on my devices now, both on internal and SD. Saving on SD card is quite slow, but I dont think theres much to do.

ltGuillaume:
It looks nicer maybe, but to be totally honest, I think this is far from a solution: now, without even informing the user, sometimes there's lossless rotation and sometimes there's lossy rotation AND the loss of all EXIF tags. I find it hard to believe that only Bitmap.compress() _can_ circumvent the problems with writing to SD, while there's no way to be found for copying a properly EXIF-rotated file from a temporary file in applicationInfo.dataDir (which worked) to a location on the SD, don't you think? It might be 1 step more, but it's a lot cleaner.

Furthermore, I found this little issue with your latest commit:
- Rotate once and save the pic to a different file
- Then rotate again immediately after
Result: rotation is again only 90 degrees (so it looks like nothing happens, because it was already rotated 90 degrees)

tibbi:
well I tried saving EXIF on SD card files already multiple times, it just didnt work. Maybe I will find a solution someday.

tibbi:
I could copy the old exif to the new one when saving to a different file on the internal storage tho

ltGuillaume:
Errr, that's exactly what I already proposed before :-P

tibbi:
Ive actually tested it already and it seems to be working, guess that inputstream copying copied the exif too

ltGuillaume:
Great! I tried that, too, but the only thing I bumped into was that "renameFile" didn't move from Internal storage to SD card, that's why I asked what fun that can do that I could use.

PR #435, opened 2017-11-03
Working EXIF lossless on Internal & SD

ltGuillaume:
Tested and works on both Internal Storage and External SD cards

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/gallery/activities/ViewPagerActivity.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/gallery/activities/ViewPagerActivity.kt
@@ -46,8 +46,9 @@ import com.simplemobiletools.gallery.models.Medium
 import kotlinx.android.synthetic.main.activity_medium.*
 import java.io.File
 import java.io.FileInputStream
-import java.io.FileNotFoundException
 import java.io.FileOutputStream
+import java.io.InputStream
+import java.io.OutputStream
 import java.util.*
 
 class ViewPagerActivity : SimpleActivity(), ViewPager.OnPageChangeListener, ViewPagerFragment.FragmentListener {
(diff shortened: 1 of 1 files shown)


app/src/main/kotlin/com/simplemobiletools/gallery/activities/ViewPagerActivity.kt
```diff
@@ -454,52 +455,60 @@ class ViewPagerActivity : SimpleActivity(), ViewPager.OnPageChangeListener, View
...
+                        } else {
+                            saveFile(tmpFile, bitmap, it as FileOutputStream)
+                        }
+
+                        if (tmpFile.length() > 0 && selectedFile.exists()) {
+                            deleteFile(selectedFile) {}
+                        }
+                        copyFile(tmpFile, selectedFile)
+                        scanPath(selectedFile.absolutePath) {}
+                        deleteFile(tmpFile) {}  // Not necessary?
```
tibbi: guess you can create a "finally" block after the 2 catch blocks and replace 3 "deleteFile(tmpFile)" with 1 in it

app/src/main/kotlin/com/simplemobiletools/gallery/activities/ViewPagerActivity.kt
```diff
@@ -454,52 +455,60 @@ class ViewPagerActivity : SimpleActivity(), ViewPager.OnPageChangeListener, View
...
-                deleteFile(destination) {
-                    renameFile(tmpFile, destination) {}
-                }
-
-                toast(R.string.file_saved)
-                mRotationDegrees = 0f
-                invalidateOptionsMenu()
+            if (needsStupidWritePermissions(destination.absolutePath)) {
+                getFileDocument(destination.parent)
             }
```
tibbi: this getFileDocument block isnt needed, its not doing anything, just slowing down the copy by getting a useles filedocument

app/src/main/kotlin/com/simplemobiletools/gallery/activities/ViewPagerActivity.kt
```diff
@@ -510,34 +519,13 @@ class ViewPagerActivity : SimpleActivity(), ViewPager.OnPageChangeListener, View
...
-
-    private fun rotateFileByExif(path: String) {
-        val exif = ExifInterface(path)
-        val orientation = exif.getAttributeInt(ExifInterface.TAG_ORIENTATION, ExifInterface.ORIENTATION_NORMAL)
-        val orientationDegrees = (degreesForRotation(orientation) + mRotationDegrees) % 360
+    private fun saveRotation(input: File, output: File) {
+        copyFile(input, output)
+        val exif = ExifInterface(output.absolutePath)
+        var orientation = exif.getAttributeInt(ExifInterface.TAG_ORIENTATION, ExifInterface.ORIENTATION_NORMAL)
+        var orientationDegrees = (degreesForRotation(orientation) + mRotationDegrees) % 360
```
tibbi: you can make these 2 variables "val" instead of "var", it is showing a hint at android studio

tibbi:
seems to be working fine on both of my device Android 6 and 7 with SD cards too, nice job!

app/src/main/kotlin/com/simplemobiletools/gallery/activities/ViewPagerActivity.kt
ltGuillaume: Aaah yeah, true. Used to be vars, didn't change.

I'm still used to Eclipse, and all the Android Studio hints and helps are still really annoying for me. Thanks for showing me the ropes!

app/src/main/kotlin/com/simplemobiletools/gallery/activities/ViewPagerActivity.kt
ltGuillaume: I'm not sure, but maybe ```if (needsStupidWritePermissions(destination.absolutePath)) {
  getFileDocument(destination.parent)
}``` is actually still necessary! I just got the following:
- Installed Gallery anew
- Installed a new SD card
- Tried to rotate an save to SD card. I got a toast message "invalid URI".
- It turned out that Simple Gallery didn't have the SD card write permissions yet, _and didn't ask for it either_.
- In another operation (I believe deleting a photo), I _did_ get the permission dialog, had to go to Android's File Manager and select the SD card.
- Afterwards, the exact same rotate operation that previously failed, worked flawlessly.

tibbi: it wouldnt be solved properly by this block. Adding a handleSAFDialog(selectedFile) {...} after creating "selectedFile" and wrapping everything after it in the new brackets should do it.

PR #437, opened 2017-11-03
Rotation button fix

ltGuillaume:
When rotating is initiated, show rotate button permanently to facilitate further rotation. Also, prevent accidental click on Delete when the option "replace share with rotate" is enabled.

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/gallery/activities/ViewPagerActivity.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/gallery/activities/ViewPagerActivity.kt
@@ -236,6 +236,11 @@ class ViewPagerActivity : SimpleActivity(), ViewPager.OnPageChangeListener, View
             findItem(R.id.menu_share_1).isVisible = !config.replaceShare
             findItem(R.id.menu_share_2).isVisible = config.replaceShare
             findItem(R.id.menu_rotate).isVisible = currentMedium.isImage()
+            findItem(R.id.menu_rotate).setShowAsAction(
+                if (mRotationDegrees !== 0f)
+                    MenuItem.SHOW_AS_ACTION_ALWAYS
+                else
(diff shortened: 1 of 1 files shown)


tibbi:
ye seems to work fine, "Save as" is still higher priority and so it is visible. Just move it as the last item at menu.apply and add some brackets, so it looks ok. 

While you are at rotating, I also want to start showing a degree picker at pressing the rotate button, so for example for rotating by 270 degrees you dont have to press it three times, just press "Rotate" and "Rotate left". It was already implemented some time ago as kind of submenu items, it would be good to have it back. It was removed at https://github.com/SimpleMobileTools/Simple-Gallery/commit/3c3740731564b47b1be0e9c608b7852a2cc2eef2#diff-c3051abca494a588aa5e78fff5374477 . Thanks

ltGuillaume:
I think that's unwise, since it'll mean that you 1) always have to press twice, and 2) you have to search for the right menu item AND anticipate on how many degrees you want to rotate it. Before you've done that you have already clicked the button (that's at a fixed position) until it's in the right orientation, way better.

ltGuillaume:
Fixed... unless you'd like the last `)` on a new line, too :-P

tibbi:
okay, leave it then. Since the rotate button will be more visible now, it should be good enough. Thanks

tibbi:
so I split the button afterall as people expect it to work like that and it also saves time

ltGuillaume:
I don't like it for when the option to always show the rotate action button is enabled, but it's your choice of course ;-)

ltGuillaume:
Are you open to making this split button optional (or off) when the share button is replaced by the rotate button? I've seen too many ppl. now - including my poort mom ;) - who were not happy at all with it. Direct visual feedback (instant rotation) is so much easier for most ppl.

tibbi:
I want to keep it as it is for now, not add too many unnecessary options.

PR #443, opened 2017-11-07
Adding handleSAFDialog for EXIF rotation

ltGuillaume:


tibbi:
have you tested if it works properly? I just wrote it from the top of my head

ltGuillaume:
Yes, I tested it. The only issue I had is more global: it tends not to refresh the folder properly, so it "looks like" the photo is gone. I have the same thing on this device in some other situations, though: sometimes a newly created photo _is_ shown as the folder preview, but once I enter the folder, that new photo is nowhere to be found. Even pull-to-refresh does not make the photo appear. Most of the times, going back and then entering the folder again fixes it, but in the case of a freshly rotated photo after the handleSAFDialog, it simply won't appear even when re-entering the folder or even when closing and re-opening Simple Gallery. Stupid problem, not enough details and hard to isolate and reproduce... sorry.

tibbi:
ok thats good enough, thanks

PR #451, opened 2017-11-10
Update strings.xml

gcalzo:
I have removed a backslash (before a "single quote") into "app_long_description" because it's shown on both my phone and on the web!! :-?


tibbi:
That way I cannot compile the app :p I will try removing the quotes from playstore descriptions someday, nobody reads it anyway.

gcalzo:
...I read it... :-D

PR #481, opened 2017-11-30
Translation update

fricyo:


tibbi:
@fricyo I wrote a guide how to set up a file watcher which will send you an email whenever a file changes, you can check it out at https://github.com/SimpleMobileTools/General-Discussion#how-can-i-get-notified-if-a-translated-file-changes . That way you won't have to keep checking files over and over.

fricyo:
Thank you for your guide.

PR #483, opened 2017-12-02
Update strings.xml

trubitsyn:


trubitsyn:
@tibbi Feel free to ping me after adding new strings to Simple apps.

tibbi:
I cannot really ping every translator, that would take a lot of time :/ I already wanted to find some way of automatic notifications if a file changes, something like https://app.github-file-watcher.com/ . I will do more research someday, maybe this file-watcher will work fine.

tibbi:
@trubitsyn I wrote a guide how to set up a file watcher which will send you an email whenever a file changes, you can check it out at https://github.com/SimpleMobileTools/General-Discussion#how-can-i-get-notified-if-a-translated-file-changes . That way you won't have to keep checking files over and over.

PR #499, opened 2017-12-13
Created norwegian translation (nb)

FTno:
Norwegian translation (nb) for Simple Gallery

By the way in this string in the original language file:

<string name="hide_folder_description">This function hides the folder by adding a \'.nomedia\' file into it, it will hide all subfolders too. You can see them by toggling the \'Show hidden folders\' option in Settings. Continue?</string>

Should not the quoted text 'Show hidden folders' be 'Show hidden media' instead.
Cannot find  'Show hidden folders' anywhere.


tibbi:
right, thanks. I will make it "show hidden items" so I can reuse it at multiple apps :)

PR #518, opened 2017-12-22
Create strings.xml

rex07:
Add ar (Arabic) translation


tibbi:
@rex07 if you could split the long app description in 5 lines like at the english version, that would be appreciated. Handling right-to-left strings is pretty confusing to me :) Thanks

PR #551, opened 2018-01-02
Update DE

Phoenix1747:
push for commit 4f4fb8bc674ede80f134b01e4a150c2a970fe2c8

### DIFF
--- a/app/src/main/res/values-de/strings.xml
+++ b/app/src/main/res/values-de/strings.xml
@@ -7,7 +7,7 @@
     <string name="hidden">(versteckt)</string>
     <string name="pin_folder">Ordner anheften</string>
     <string name="unpin_folder">Ordner loslösen</string>
-    <string name="pin_to_the_top">Pin to the top</string>
+    <string name="pin_to_the_top">An den Start anheften</string>
     <string name="show_all">Ansicht: Alle Medien</string>
     <string name="all_folders">Alle Medien</string>
     <string name="folder_view">Ansicht: Als Ordner</string>


tibbi:
Hey @Phoenix1747 a guy got confused in the translation for too similar words used for "Hide" and "Exclude". He says that for excluding it would be better to use "ausschließen" instead of "ausblenden" to make the difference clearer, does it make sense? Thanks

Phoenix1747:
It's a bit confusing with the two different hiding options, I admit. But I guess it makes totally sense, I just checked it in the app. Just to make it clear there are two options (in German)...
1. `Ordner verstecken` which hides the selected folder globally (via the .nomedia file).
2. `Ordner ausblenden` which hides the selected folder in-app only.
In the context of hiding the folder globally it would make more sense to use the German equivalent of `Hide folder` instead of `Exclude folder`, at least in my opinion this makes it more clear. But what did the guy talk about exactly?

tibbi:
he said:
The upper right menu gives two seperate options:
- hide only functionality (uses .nomedia and can be shown temporarily) (in screenshot as "Ordner verstecken")
- exlude folder by "blacklisting" it (no .nomedia required) (in screenshot as "Ordner ausblenden")

I have chosen the wrong one up to this day, not realizing that there are two different options.
There is a small translation error, which isnt wrong, but only unfitting.
Both translations are synonyms and both mean "to hide". It would be better to use "Ordner ausschließen" to exclude a folder. This is a better translation and should fix the trouble.

Regarding the second bug:
In the settings there are these three options:
- manage included folders (translated as "Einbezogene Ordner verwalten")
(Well translated)

- manage excluded folders (translated as "Ausgeblendete Ordner verwalten")
Better translation: "Ausgeschlossene Ordner verwalten"
(Grammar is correct)

Phoenix1747:
Ok, right now I understand. But IMHO you could really use both options here. However there is like more difference between two menu options `Ordner ausschließen` and `Ordner verstecken` than between the current two `Ordner ausblenden` and `Ordner verstecken`. Also `Ordner ausschließen` would be the exact literal translation.
But yeah, I think you can use both so choose what you like more :P

tibbi:
I dont know, I cannot decide. But if you update the string, update "Manage excluded folders" too, or wherever else it is. Thanks

Phoenix1747:
Ok then, I'll just update this 👍

PR #561, opened 2018-01-07
Fix German translation

Phoenix1747:
Regarding PR #551


tibbi:
okay cool, thanks

tibbi:
@Phoenix1747 another guy said that the translation of Manage Included Folders placeholder at https://github.com/SimpleMobileTools/Simple-Gallery/blob/master/app/src/main/res/values-de/strings.xml#L54 is a bit misleading. It is now used in a way that you add folders there only if some media in the folders isnt recognized, but it sounds like it was supposed to be used if the folder itself wasnt recognized. Not sure if it makes sense, here is his text. He also suggested an edit to the english version, but I dont want to change that.

But let me say, that description of "Manage Included Folders" is kind of misleading. It says:
"If you have some folders which contain media, but were not recognized by the app, you can add them manually here."

This sounds like folders might be not recognized. But in my case, the folders were recognized correctly. Instead of it the media files being inside the folders were not recognized.

The german translation now strengenths this misunderstanding:
"Solltest du weitere Medienordner haben, die von der App nicht erkannt wurden, kannst du diese hier manuell hinzufügen."

The translation back in english means:
"If you have more media folders, which were not recognized..."

So I would like to make another suggestion for the description (english):
"If you have some media files, which were not recognized correctly, you can add their folders manually here"

Phoenix1747:
Ok, I didn't translate this string originally, but this is true, it's indeed misleading; fixed it.

tibbi:
alright, thanks

Issue #578, opened 2018-01-15
App closes with no media or all folders are excluded when switching to "show all folders content"


gh123man:
If you exclude all media containing folders, or have no media on the device, switching to "show all folders content" will exit the media activity immediately. Upon subsequent launches it is impossible to change the setting back from within the app.

PR #579, opened 2018-01-15
fixed bug where app will close if all media is missing or excluded

gh123man:
This is a fix for #578 and possibly #574 
I am not sure of the context/repro for #574


tibbi:
ye nice one again, thanks

PR #610, opened 2018-01-29
Dutch + 2 remarks

ltGuillaume:
1. When "Replace zoomable images" is enabled, I can actually zoom less than when it's disabled
2. One finger zoom somehow puts the image upside down after some zooming in and out. Perhaps EXIF-tag or "rotate according to device orientation" related?

(Not the right place, I know)

### DIFF
--- a/app/src/main/res/values-nl/strings.xml
+++ b/app/src/main/res/values-nl/strings.xml
@@ -133,13 +133,13 @@
     <string name="manage_extended_details">Uitgebreide informatie</string>
     <string name="one_finger_zoom">Met één vinger zoomen in volledig scherm</string>
     <string name="allow_instant_change">Direct naar vorige/volgende door op de zijkanten van het scherm te tikken</string>
-    <string name="replace_zoomable_images">Replace deep zoomable images with better quality ones</string>
-    <string name="hide_extended_details">Hide extended details when status bar is hidden</string>
+    <string name="replace_zoomable_images">Betere kwaliteit hanteren bij ver inzoomen</string>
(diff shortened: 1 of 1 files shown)


tibbi:
1. sounds right, no?
2. does it happen to every image, or only the ones coming from the same source? Like the same camera app?

ltGuillaume:
1. Then I think I don't understand the feature? Shouldn't I be able to zoom further with that feature turned on? I can only zoom in a fraction of what I was used to, unless I turn the feature off.
2. Yes, no matter what the image source is, _only happens when both "replace zoomable images" and "one finger zoom" are enabled_

tibbi:
1. no, by default Im using the library thats purpose is zooming in high resolution images. With it you can zoom in deeply, but the image quality is lower. If it is turned on, Im using the library with better quality.
2. weird, works fine to me. That one finger zoom is actually enabled no matter if you have it enabled or not, Ill tweak it in the future

ltGuillaume:
1. All right, then I'd say the description "deep zoom" is a bit confusing, because it actually prevents you from "deep zooming", wouldn't you agree?
2. I tried the one finger zoom (double tap, then swipe down/up) and it's only enabled when the option is.

tibbi:
I guess you are misunderstanding the first point and think it is vice versa. When it is enabled, images have better quality. When it is disabled, you have deep zoom. Thats what the english version describes too.

ltGuillaume:
Yes, sorry, I misunderstood. New pull request submitted.

PR #619, opened 2018-01-30
Set unique applicationIdSuffix in debug builds

trubitsyn:


tibbi:
append the launcher name with _debug or something too, else this will be a mess. Calendar has a debug string for leak canary. you can copy it. Thanks

trubitsyn:
I'll have to create an override for launcher name in `src/debug/res/values/strings.xml`, are you OK with this?

tibbi:
sure, thats what I meant

trubitsyn:
Seems like I'll have to create that string resource for `every` language in `debug`, it takes localized strings from `main` source set otherwise. Do you know any better approach?

tibbi:
not sure. I tried a couple things, none of them seemed to work. Just add it in the english version then, at the default "values". Or in other language if you use an other one.

trubitsyn:
Don't like the current solution. Appending "_debug" or any other string is almost indistinguishable. On a Pixel emulator this suffix wouldn't even be displayed for e.g. Music Player because the launcher cuts the name. I think changing icon's color would be the best solution: the original name is preserved and the apps being developed are *very* distinguishable. I'll try to find a Gradle plugin that dynamically generates app icons to avoid cluttering the app with PNGs that make sense only in debug mode.

tibbi:
well it certainly needs to be distinguished somehow, I dont know what is the easiest way

tibbi:
I wouldnt change the icon color tho, my device will look terrible. Name would still be the best.

trubitsyn:
@tibbi added the suffix to all launcher names.

tibbi:
ok thx

PR #682, opened 2018-03-01
Updated German FAQ

kniddl:
"That will make only the selected folders visible, as both excluding and including are recursive and if a folder is both excluded and included, it will show up."
What did you mean with the last bit of the sentence (it will show up)? The folder will nevertheless show up, I understood it as that, or that there will be a visible warning?
If there is a visible warning, it would be as follows: "...einbezogen wird, wird eine Warnung angezeigt."


tibbi:
no theres no visible warning. It just means that the folder will be visible

kniddl:
Okay, then my translation fits.

PR #696, opened 2018-03-11
Update strings.xml

matiasblomqvist:
More Finnish translations. Lots of stuff wasn't in the file, so I added them from:
https://github.com/SimpleMobileTools/Simple-Commons/tree/master/commons/src/main/res
like it says in the bottom. I hope I used the correct tags.


tibbi:
you shouldnt add any new strings in any file, just translate the Commons files themselves at https://github.com/SimpleMobileTools/Simple-Commons/blob/master/commons/src/main/res/values-fi/strings.xml . Click on the link and edit it the same way by pressing the pencil. Please create a new pull request to make it cleaner. One pull request should be here, the second one at the Commons repository. Thanks

PR #716, opened 2018-03-24
Update russian strings

mrvak:
It can contain some mistakes, but it will be better, than Goggle_translate-like translating. (maybe it has done by a human, who dont speak russian as a first language)


tibbi:
okay, we will see if anyone will have some comments about it, I cannot decide :) Thanks
PS: another group of strings reused by multiple apps is at https://github.com/SimpleMobileTools/Simple-Commons/blob/master/commons/src/main/res/values-ru/strings.xml if you are interested.

PR #718, opened 2018-03-25
Added Greek translations

sotcha:


tibbi:
Thanks :) Btw, feel free translating app name and launcher name too.

PR #748, opened 2018-04-15
Delete values-lt

mobtechpd:


tibbi:
I moved the file in the proper place, didnt notice the issue before

PR #836, opened 2018-07-07
Translate missing strings in german

CanRau:
Translated:
```
force_portrait
force_landscape
use_default_orientation
```

I'm new to simple mobile tools and don't yet know what words best fit the expectation/experience, so let me know if any german speaker has suggestions or of cause I wont cry if someone else sees a better fit ;-)


tibbi:
I dont know a better word either :) If some german speaker thinks it isnt a perfect fit, he will suggest a better one. Thanks!

PR #899, opened 2018-08-09
Dutch

ltGuillaume:
So, how does this compare to "Replace deep zoomable images with better quality ones"? The option is reversed and now enabling this option is "at the cost of image quality"?


tibbi:
it is always good quality now, so just the deep zooming it toggled. On some devices it causes lags.

ltGuillaume:
Cool! Thx for the info.

PR #995, opened 2018-10-12
Update German FAQs

kniddl:
"Du" is the personal form of address and "Sie" is the honorific. Which do you prefer? I saw that in French the honorific form of address was chosen but in German not.

### DIFF
--- a/app/src/main/res/values-de/strings.xml
+++ b/app/src/main/res/values-de/strings.xml
@@ -37,7 +37,7 @@
     <string name="images">Bilder</string>
     <string name="videos">Videos</string>
     <string name="gifs">GIFs</string>
-    <string name="raw_images">RAW images</string>
+    <string name="raw_images">RAW-Bilder</string>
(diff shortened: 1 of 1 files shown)


tibbi:
I dont know, just make it consistent in the whole file

kniddl:
A clear stance would be easier since I am not the only translator. As long as people are interpreting the you differently there will always be inconsistencies. It's the choice between casual or polite and I can't make the decision for, sorry.

tibbi:
Im not from Germany, I dont know what is the usual way for doing it. Just check other strings how is it done, no need to change it.

tibbi:
Im using the more polite way, but or example in Hungary they use the casual version more often. Just leave the German strings as they are then. Thanks

kniddl:
Ok, I will just watch to have it consistent within an app and not across all.  And I will use the polite form in my translation then. Thanks for taking your time and thanks for the apps; they are very useful 😄.

Issue #505, opened 2017-12-16
[Feature request] Remember last video position and return to it when opening the video again


leafcutterant:
The title basically tells it all. It would be very useful.

mathevs:
Can i work on this? Is it still free?

tibbi:
ye

tibbi:
will be added in 5.1.3

PR #1038, opened 2018-10-30
#505 remembering last video position

mathevs:
#505 adding feature. Saving last video position and storing it to the config. After starting video again it will continue from last remembered position.


tibbi:
make it optional and disabled by default please, plus fix the code formatting (spaces) and conflicts. Thanks

mathevs:
It is ok now, but some strings need to be translated.

tibbi:
you have to put the string in every language file anyway, even if obviously in english

mathevs:
I put this string to every language string.xml.

mathevs:
I already moved it to the right position https://github.com/SimpleMobileTools/Simple-Gallery/pull/1038/commits/494e71e5e34a88255033e449cec47ce407ddeba0

tibbi:
aha I didnt see it. Thanks, looks good, it just doesnt work properly if there are multiple video files near each other. Ill fix it.

Issue #159, opened 2017-02-07
add video length on video thumbnail


hussamjan:
make it easier to know how long a video is without need to open it.

dslivka1:
Hi @tibbi , It is open ? I cloud do it.

tibbi:
ye it is open, but make it optional and disabled by default

tibbi:
this has been added already

PR #1073, opened 2018-11-09
#159 adding option to show video length in thumbnail

dslivka1:
Adding option to show video length in thumbnail of the video. The option is disabled by default and it can be enabled in the settings.


tibbi:
looks good, thanks

PR #1129, opened 2018-12-12
Add option to play videos externally

TheRandomLabs:
After scrolling from another picture/video to a video then clicking on the play button, Simple Gallery by default uses ExoPlayer. I've implemented an option that allows the user to change this behavior.


tibbi:
thanks, but I dont want to add such redundant checks, nobody requested it so far. There is already an "Open with" button in case someone wants to use another app.

TheRandomLabs:
Fair enough. What about changing the open with button to not force an app
chooser?

Le jeu. 13 déc. 2018 à 02:19, Tibor Kaputa <notifications@github.com> a
écrit :

tibbi:
I want to keep it as it is for now, it might be changed in the future

TheRandomLabs:
Alright, thanks!

tibbi:
so it turns out that I will use that way of playing videos indeed, but Ill make it mandatory, not optional. It would fix a few glitches that I couldnt solve other way.

TheRandomLabs:
Cool, great to hear!

PR #1166, opened 2018-12-30
Update Hungarian strings.xml

hevesij:


tibbi:
please keep the short description up to 80 characters long

PR #1195, opened 2019-01-09
Update strings.xml

smarquespt:


tibbi:
again, the short desc has to be less than 80 chars

PR #1200, opened 2019-01-12
Dutch FAQ

ltGuillaume:
Doesn't faq_9_text refer to a setting that's no longer present?


tibbi:
ye it should be modified, Ill get there :) Thanks

PR #1238, opened 2019-01-28
Update strings.xml

Solatec:


tibbi:
@Solatec here too, please make the short app description at the bottom of the file shorter than 80 characters. Thanks

PR #1282, opened 2019-02-22
Updated fr (French) translation

unknowtech:
Commit made via [Stringlate](https://lonamiwebs.github.io/stringlate/)


unknowtech:
I made too many mistakes, it's not clear.

PR #1326, opened 2019-03-10
Update strings.xml

spkprs:
google play descriptions


tibbi:
heh nice, I didnt expect someone translating it that early. Make the short app description shorter please. The max allowed length is 80 characters, the current one is 118. Feel free leaving out some parts, I know that squeezing it into 80 is difficult in some languages.

PR #1329, opened 2019-03-11
Dutch

ltGuillaume:


tibbi:
make the short app description max 80 chars please. Leave out some part of it if you cannot squeeze it. Thanks

PR #1332, opened 2019-03-11
Update strings.xml

Solatec:


tibbi:
oh that is a lot of cut data, 43/80 chars. You could add some things back, if they fit in :)

PR #1338, opened 2019-03-14
Trim german short description to 80 chars

mklopp:
I guess &amp; counts as one character.


tibbi:
hah ye, thanks. 80/80

PR #1340, opened 2019-03-15
Update italian translation to 064bf7a

ema-pe:
This pr needs some review, if no one will do that I'll review the translations in the next days.


tibbi:
well we wont review it, I doubt any other volunteer will. I guess you can confirm it.

ema-pe:
It seems ok

tibbi:
Make the short app description shorter please, it can have max 80 characters. It has 88 now. Thanks

PR #1358, opened 2019-03-29
Update Danish translation

chreddy:
Removed the previous pull request, as I ended up translating the entire FAQ section into Danish.
Also I realized that my translation of the app_short_description string was a few characters too long. Sadly I haven't been able to come up with a new translation for this string, without leaving out a part of the description. So I'll leave it as is for now.


tibbi:
Hey, just leave out a part of the description, if you cannot squeeze it. It know that it is hard to fit into the 80 chars, many language translators have left out parts of it. Thanks

PR #1360, opened 2019-03-31
Switch from using one single stored position to one position per file

centic9:
This allows to watch multiple files partially where all of them will have their position saved.

This also removes the current saved position/path settings, no migration done as it was always only remembered for one file anyway and "lost" as soon as you opened any other file.


tibbi:
looks good, just remove the System logs, Im not using them. And concat strings the kotlin way, like "$LAST_VIDEO_POSITION_PREFIX${path.toLowerCase()}". Thanks

centic9:
Sure, done, I also fixed a few recently introduced problems in the Danish translation that caused a bulid-failure for me.

tibbi:
ok, it looks good now, thanks

PR #1382, opened 2019-04-08
Updated Russian language

solokot:


tibbi:
by removing something for shortening I meant removing from the "Photo Manager &amp; Editor" part, not from the app name :) Translate for example "Simple Gallery Pro: Photo Manager". Thanks

PR #1402, opened 2019-04-29
FIX #1368 Move video duration to top right

### LINKED ISSUE #1368: Suggestion - Move video duration to top right
If the filename is long enough, it'll overlap the video duration. If it was moved to the top right, things would look a bit better.
Screenshot_20190402-200008

OlgaMorales:


tibbi:
this moves the filename, not the duration. Please at least test your code before submitting.

OlgaMorales:
It was a mistake, I'm sorry.
you fixed it , nice!

PR #1454, opened 2019-06-07
Update strings.xml

Solatec:
I thought they had already translated all ...


tibbi:
looks good now, thanks

PR #1466, opened 2019-06-19
Update strings.xml

32bituser:
It should to change the name of this switch to more accurately reflect its essence, since it is the switching between grouping by some of the selected criteria (date, type, etc.), or by folders. Variant: "View by criteria / by folder".

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -9,9 +9,9 @@
     <string name="pin_folder">Pin folder</string>
     <string name="unpin_folder">Unpin folder</string>
     <string name="pin_to_the_top">Pin to the top</string>
-    <string name="show_all">Show all folders content</string>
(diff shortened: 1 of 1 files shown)


tibbi:
you seem to have messed up some strings, it isnt the intended behaviour

32bituser:
I don't think so. Look at the option "Show all folders content" (by the way, it been messed with "Temporary show hidden" very often, at the Russian translation at least):
Scr1
After we set this option we really see all folders content, but all files now are grouped by some criteria (by date by default) and we can change this criteria (filetype,extention etc.) using the option "Group by" (screenshot 2 pos.1). Then we can return an original "group by folders" view using the option "Switch to folder view". So, I think that "Group by criteria\Group by folder" is more clear than "Show all folders content\Switch to folder view". There may be some discussions about the English terms, but the current Russian translation is definitely not the best!
Scr2

Issue #1487, opened 2019-06-27
[Feature Request] Show deleting progress in the notification bar


CatKasha:
After deleting or moving files from the SD card, the deletion process in SD card takes some time, and in the notification bar there is no indication of this operation

PR #1560, opened 2019-08-21
#1487 Show deleting progress in the notification bar

victorvicari:
Add a progressNotificationAsyncTask class and move the delete logic from main activity to doInBackground(). The notification is being created inside the AsyncTask, at onPreExecute()


tibbi:
I dont want to use it in this form as deleting has to be handled in the Commons library, since it affects multiple apps. But I might use it for inspiration when Ill work on it. Thanks

PR #1633, opened 2019-11-06
Revert "Update strings.xml"

sujebi7741:
Revert "Update strings.xml"


tibbi:
why is that?

arkar787:
Please add Network Drive (NAS, SMB) support, ( do not need Internet access in-app, just want local Network access.)

PR #1674, opened 2019-12-22
Set proper permissions on gradlew script

MariusVolkhart:


tibbi:
why is this needed?

MariusVolkhart:
On Linux/Mac, the executable flag must be set for the script to be runnable
via command line such as ./gradlew build

tibbi:
I see. I use Linux too, but not really running that script manually

PR #1717, opened 2020-01-12
Add German strings

fscholdei:
I hope the German translations have not become too long. Otherwise, Germans get along well with the following terms from English:
- Custom instead of "Benutzerdefiniert"
- Reset instead of "Zurücksetzen"

I didn't find any function for the bringToFront text in the app. Maybe foreground ("Vordergrund") would otherwise be a good translation. So I first focused on the meaning of the term.


tibbi:
cool thanks, looks ok. Ye, I know that german words can be quite long. Anyway, there are a few more untranslated strings in the given file, spread here and there, if you wanted to translate them too :)

PR #1718, opened 2020-01-13
Add German strings

fscholdei:
I had translated in pull request #1717 starting from the marked line. Hopefully I haven't missed anything now. For the deleted punctuation marks I had followed the English strings.xml.


tibbi:
cool, thanks :) The dot after the english short description is missing just because it didnt fit into the 80 chars limit, it is already 80 characters without it. It doesnt matter if the dot is there, you can leave it as it is.

PR #1728, opened 2020-01-28
Loop videos without gaps between playback

ForgottenUmbrella:
As GIFs continue to be replaced by videos, it becomes increasingly important to support gapless video loops.
ExoPlayer 2.5 brought upstream support for seamless video looping functionality through prebuffering. This behaviour is preferable to the current implementation, which seeks to the start of the video and has a noticeable, if minor, delay.

Note: a minor downside of using ExoPlayer's implementation is that certain broken videos which previously played in Simple Gallery will no longer start, due to an as-of-yet unfixed bug (google/ExoPlayer#6459). Such videos have an audio track of only a few samples (less than a fifth of a second), and playback could be re-enabled by disabling video loops (and falling back on the seek method) or disabling the audio track itself. However, ExoPlayer unfortunately does not provide any way to check the duration of the audio track in order to do this automatically, so there is no workaround. I figure that this shouldn't be a deal-breaker, since broken videos aren't expected to be supported to begin with.


tibbi:
seems like it didnt break anything during my testing, so if it improved anything, it should be good :) Thanks

PR #1755, opened 2020-02-21
Not to refresh and show all media even query text is not empty.

uhm0311:
It's very uncomfortable when I search image in directory that contains over 2,000 images. MediaActivity shows all images even though query text is not empty about 1~2 seconds after onCreate called. Please consider merge this PR.


tibbi:
looks good, thanks

PR #1761, opened 2020-02-24
Remaining strings translated to Polish

kryger:
I stumbled upon an untranslated settings entry and looked it up here on github, decided to go through the entire file. Please let me know if any of the translations cannot be accepted because of technical reasons (too long to fit the UI etc.)

Source: I'm a native speaker.


tibbi:
looks good, should be ok. Thanks :)

PR #1766, opened 2020-02-29
The appid for foss is already changed

licaon-kter:
...no need to change the name too.

This breaks autoupdate in F-Droid atm: https://gitlab.com/fdroid/fdroidserver/issues/733


tibbi:
I actually removed that line together with applicationIdSuffix

PR #1850, opened 2020-05-31
Update pronoun and fix a few missing translations for zh_CN.

lovenemesis:
As titled.

Thank you!


tibbi:
nice, thanks for the translations :)

PR #1857, opened 2020-06-09
Setting EXIF DateTaken from file name and "inverse term" search

jkumeboshi:
As foreword, sorry for submitting a code that is far from your standards, but I'm not used with all Android's limits on writing on SD cards, so I wrote something that just seems to work for me using already existing functions that I'm far from understand how to properly call.

I've added 2 new feature: 
1) reverse search term, to search all items NOT containing a search term, just prefix the term with !

2) restoring date taken EXIF info from file name containg a timestamp (many Xiaomi save file format and WhatsApp image files: "IMG_YYYYMMMDD_hhmmss.jpg" / "IMG_YYYYMMMDD-WAXXXXXXX.jpg" ). Also the file's lastModified is set to same timestamp.

Please fell free to reject the request to re-implement any of above feature by yourself, or modify it.

Regards and thanks a lot for sharing your great work!


tibbi:
right, this really doesnt look like the rest of the code used in the app, and the string has a typo too. While the implementation is just weird, I will try implementing something like that in the future. Thanks

PR #1903, opened 2020-09-01
Replaced context!! activity!! with requireContext() requireActivity() for null safety

GeorgCantor:


tibbi:
I dont really see the benefits of this, nor I get any related crashes. Thanks though

PR #1907, opened 2020-09-11
Use Core KTX extensions.

Isira-Seneviratne:
* Use [extension functions](https://developer.android.com/kotlin/ktx/extensions-list) included in the Core KTX library.
* Add extensions that call [`ShortcutManagerCompat`](https://developer.android.com/reference/androidx/core/content/pm/ShortcutManagerCompat).

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -70,6 +70,9 @@ android {
         sourceCompatibility JavaVersion.VERSION_1_8
         targetCompatibility JavaVersion.VERSION_1_8
     }
+    kotlinOptions {
+        jvmTarget = JavaVersion.VERSION_1_8.toString()
+    }
 
     packagingOptions {
         exclude 'META-INF/library_release.kotlin_module'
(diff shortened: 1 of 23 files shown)


tibbi:
what is the benefit of this?

Isira-Seneviratne:
> what is the benefit of this?

It makes the code more Kotlin-idiomatic, and some of the extension functions make use of compatibility mechanisms, e.g. the `getSystemService()` extension calls [`ContextCompat.getSystemService()`](https://developer.android.com/reference/androidx/core/content/ContextCompat#getSystemService(android.content.Context,%20java.lang.Class%3CT%3E)) and the `parseAsHtml()` extension calls [`HtmlCompat.fromHtml()`](https://developer.android.com/reference/androidx/core/text/HtmlCompat#fromHtml(java.lang.String,%20int,%20android.text.Html.ImageGetter,%20android.text.Html.TagHandler)).

tibbi:
well that sadly isn't worth the change, thanks though

PR #1909, opened 2020-09-11
Remove calls to MenuItemCompat.setOnActionExpandListener().

Isira-Seneviratne:
`MenuItemCompat.setOnActionExpandListener()` has been deprecated, and its documentation recommends calling `MenuItem#setOnActionExpandListener()` directly.


tibbi:
it used to be like that, but it was crashing. Anyway, I dont want to risk such changes just because something is deprecated, if it won't bring any benefits to the users.

Isira-Seneviratne:
The current method simply calls `MenuItem#setOnActionExpandListener()` directly.

PR #1915, opened 2020-09-18
Use ShortcutManagerCompat.

Isira-Seneviratne:
Use [`ShortcutManagerCompat`](https://developer.android.com/reference/androidx/core/content/pm/ShortcutManagerCompat)'s `isRequestPinShortcutSupported()` and `requestPinShortcut()` methods to create shortcuts even on API levels below 25.


tibbi:
I dont want to risk breaking anything, I just checked and literally 99% of installs is from Api 25+

Isira-Seneviratne:
> I dont want to risk breaking anything, I just checked and literally 99% of installs is from Api 25+

Then why not bump the minSdk version?

tibbi:
because I have no reason for bumping it

PR #1916, opened 2020-09-19
Use coroutines.

Isira-Seneviratne:
Use Kotlin's [coroutines](https://kotlinlang.org/docs/reference/coroutines-overview.html) to improve performance, particularly when scrolling through folders containing lots of images.


tibbi:
Thanks for the effort, but it doesnt really improve the performance to me, nor I see why would it. Changing this amount of code in 27 files at once is way too risky, plus you removed some try/catch blocks for no reason.

Isira-Seneviratne:
> Thanks for the effort, but it doesnt really improve the performance to me, nor I see why would it. Changing this amount of code in 27 files at once is way too risky, plus you removed some try/catch blocks for no reason.

Exception handling isn't mandatory in Kotlin, and those blocks weren't logging the exception or doing anything else.

tibbi:
all try/catch blocks were there for a reason, I saw crashes there in crashlogs

Isira-Seneviratne:
> all try/catch blocks were there for a reason, I saw crashes there in crashlogs

Oh, my bad

Isira-Seneviratne:
How about if I broke it up into smaller changes?

tibbi:
well if they will be really smaller ones, we could try. I want to release a new version with few changes too, so it wont be all released together.

PR #1917, opened 2020-09-20
Rewrite GetMediaAsynctask as a suspend function.

Isira-Seneviratne:
Starting with API level 30, [`AsyncTask`](https://developer.android.com/reference/android/os/AsyncTask) has been deprecated, with the documentation recommending that Java or Kotlin's concurrency utilities be used instead.


tibbi:
ye this looks ok, so the app can be published with this, right? Then we add the rest.

app/src/main/kotlin/com/simplemobiletools/gallery/pro/activities/MediaActivity.kt
```diff
@@ -548,41 +552,39 @@ class MediaActivity : SimpleActivity(), MediaOperationsListener {
...
-        } else {
-            getCachedMedia(mPath, mIsGetVideoIntent, mIsGetImageIntent) {
-                if (it.isEmpty()) {
-                    runOnUiThread {
-                        media_refresh_layout.isRefreshing = true
+        lifecycleScope.launch {
+            if (mLoadedInitialPhotos) {
+                startCoroutine()
+            } else {
+                getCachedMedia(mPath, mIsGetVideoIntent, mIsGetImageIntent) {
```
tibbi, line 559: this is wrong, always returns 0 files

PR #1918, opened 2020-09-21
Use suspend functions in DateTakensDao.

Isira-Seneviratne:
Add Room KTX and change the functions in `DateTakensDao` to suspend functions.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -78,6 +78,7 @@ android {
 
 ext {
     coroutines_version = '1.3.9'
+    room_version = '2.2.5'
 }
 
 dependencies {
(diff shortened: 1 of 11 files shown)


tibbi:
well okay, lets see how will it work :) Thanks

tibbi:
getCachedMedia in MediaActivity still always returns 0, there is something wrong with it. Even after opening the same folder many times, the files are not stored in the db properly.

tibbi:
Im actually getting some crashes related to accessing the database on the main thread too, so Im reverting this whole coroutine thing. It is way too unstable to be used in production.

Isira-Seneviratne:
Sorry for the late reply, I only saw this now.

Hope this helps: https://developer.android.com/kotlin/coroutines

tibbi:
I dont plan studying it, I see no benefits of switching to coroutines. This is not some school projects for testing new geeky solutions. The app is used by a lot of people, many of them even paid for it so they expect some stability.

Isira-Seneviratne:
I've worked with coroutines before on other Android projects, and I've found that they work very well.

Out of curiosity, did you use a log call to display the size returned by `getCachedMedia`, or a different mechanism? I'd like to isolate the exact cause of the issue you mentioned.

tibbi:
ye, it returned 0 and I can see the top refreshing circle too. But as mentioned previously, some DB operations are being ran on the main thread which is causing crashes, so theres quite a few issues. And the Gallery isnt right the best app for testing it, it is quite complicated.

Isira-Seneviratne:
Are you planning to revert the AsyncTask change?

tibbi:
it is already reverted

PR #1925, opened 2020-09-29
Use suspend functions in DirectoryDao.

Isira-Seneviratne:
Convert the DAO methods in `DirectoryDao` to suspend functions to prevent them being called on the main thread. Only the methods in `DirectoryDao` were converted to keep the PR as simple as possible, with the hope of converting the others in separate PRs.

Unlike the previous PRs, the methods are called using the I/O dispatcher, which is backed by a thread pool used for I/O tasks and [is recommended for use with Room](https://developer.android.com/kotlin/coroutines-adv#main-safety).

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -76,6 +76,11 @@ android {
     }
 }
 
+ext {
+    coroutines_version = '1.3.9'
+    room_version = '2.2.5'
+}
+
 dependencies {
     implementation 'com.simplemobiletools:commons:5.30.17'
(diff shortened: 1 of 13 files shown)


Isira-Seneviratne:
On another note, I converted `GetMediaAsynctask` to a suspend function in a separate branch and added logging calls to check the list contents in `getMedia()` and `getCachedMedia()`, and neither list was empty when navigating through folders containing images.

PR #1939, opened 2020-10-19
Use suspend functions in DateTakensDao.

Isira-Seneviratne:
Execute the methods in `DateTakensDao` using Kotlin coroutines (which are essentially lightweight threads) instead of creating threads manually.

**Note:** Unlike in the previous PR, the DAO functions are called using the I/O dispatcher, which is backed by a thread pool and is [recommended](https://developer.android.com/kotlin/coroutines-adv#main-safety) for use with Room, so there shouldn't be any issues related to the main thread.


tibbi:
I still dont see the benefits of this, what is it? It looks way too complicated and brings unnecessary risk.

Isira-Seneviratne:
> I still dont see the benefits of this, what is it? It looks way too complicated and brings unnecessary risk.

It allows threads to be reused instead of having to create new threads in order to update the database.

I tested out this change on my own device and didn't encounter any issues.

tibbi:
alright thanks, lets try how this works out

tibbi:
nope, it is crashing, Im reverting it

Issue #1653, opened 2019-11-27
Allow setting a folder to open by default at app start


tibbi:

v1r0x:
Great feature! It's the reason I didn't switch from google photos to this as default app

roland-kister:
Hey @tibbi , I would like to contribute.

Where would be the best place to put this setting? Into the settings of a given folder (for example: Use this folder as default folder) or the app settings.

tibbi:
Try the menu of the given folder, the app settings are already more crowded than Id like. Thanks

roland-kister:
Okay, I will get on it

tibbi:
will be added in 6.18.0

PR #1945, opened 2020-10-26
Added default folder option

roland-kister:
Added _Set as default folder_ into folder menu. When the folder is set as default, the _Unset as default folder_ is shown. _Set as default folder_ is shown in `MainActivity` menu when some other folder is already set as a default folder. Also when the app is started, the existence of default folder is checked. If it doesn't exists, it is set to `null` again.

I also added translations, but I used Google translate to get them. Any recommendation how to handle translations?

This pull request is based on #1653


tibbi:
no need to translate the strings, just add the original english strings to every languages file and wait for volunteers for translating. Make it consistent though, move them under switch_to_file_search at the first section. Also remove the semicolons in the code, they arent needed. Will take a look at actual functionality later. Thanks

app/src/main/kotlin/com/simplemobiletools/gallery/pro/helpers/Config.kt
```diff
@@ -198,6 +198,10 @@ class Config(context: Context) : BaseConfig(context) {
         get() = prefs.getInt(getDirectoryColumnsField(), getDefaultDirectoryColumnCount())
         set(dirColumnCnt) = prefs.edit().putInt(getDirectoryColumnsField(), dirColumnCnt).apply()
 
+    var defaultFolder: String?
+        get() = prefs.getString(DEFAULT_FOLDER, null)
+        set(defaultFolder) = prefs.edit().putString(DEFAULT_FOLDER, defaultFolder).apply()
```
tibbi, line 203: make it an empty string by default, do not use "null" at such places

tibbi:
this looks pretty good actually, lets see if it wont break anything :) Ill just move the strings into the proper place. Thanks

PR #1972, opened 2020-11-15
Replace AsyncTask with RxJava 3.

Isira-Seneviratne:
Use RxJava 3 instead of `AsyncTask` to carry out asynchronous operations, as `AsyncTask` is deprecated starting with API level 30.


tibbi:
and what is the benefit of switching?

Isira-Seneviratne:
> and what is the benefit of switching?

It avoids context leaks, and RxJava can also be used with Room to create reactive queries: https://medium.com/@lpereira/why-should-we-use-rxjava-on-android-c9066087c56c

zhombie:
@tibbi AsyncTask is also deprecated now, more info here: https://developer.android.com/reference/android/os/AsyncTask. But google suggests to switch to kotlin coroutines

tibbi:
many things are deprecated in Android, Im not worried about that :)

tibbi:
Ill close it as it probably isnt even tested properly, like https://github.com/SimpleMobileTools/Simple-Gallery/pull/1939

PR #1975, opened 2020-11-17
Use view binding.

Isira-Seneviratne:
Make use of view binding, as the Kotlin Synthetics feature has been deprecated in Kotlin 1.4.20 in favor of view binding.


tibbi:
Ill close it as it probably isnt even tested properly, like https://github.com/SimpleMobileTools/Simple-Gallery/pull/1939

PR #2022, opened 2020-12-22
Adds support for Bangla language

kabirnayeem99:
This PR translates 70% of the strings to Bangla, which is an Indo-Aryan language and the lingua franca of the Bengal region used by 250 million in Bangladesh and West India.


tibbi:
alright, thanks :)

PR #2052, opened 2021-01-22
Created values-eu and translated strings

avtkal:
App translated. FAQ and Playstore descriptions not translated yet.


tibbi:
nice, thanks!

Issue #1949, opened 2020-10-29
Do not turn off the screen when idle while viewing the image


williamson91:
Great app!  Thanks.  I just wanted one more function: do not turn off the screen when viewing an image.  Let it be turned off by default, but who needs to be able to turn it on.

AmeyaPhadnis-2019H1030012G:
Can I work on this, I am new to open source and would like to contribute

williamson91:
> Can I work on this, I am new to open source and would like to contribute

I think they will not mind accepting sensible pull requests.

PR #2057, opened 2021-01-24
Do not turn off the screen when idle while viewing the image #1949

tanvir-ahmod:


tibbi:
I dont really like this, it should fall asleep during viewing images. It would need an extra setting item that I dont want to add. Thanks though

m-alice:
Really sadly, i need it for watching ̶e̶r̶o̶ electronic schematics.
I will have to, among other things, implement this using the Tasker app.

Issue #1958, opened 2020-11-05
apply selection to slideshow


yepyepyep4711:
This is a feature request to make the slideshow command available after media have been selected, starting a slideshow with only those pictures/videos. At the moment, this command disappears if any media is selected. And if I want to start a slideshow with only some pictures, I need to first select them then copy them into a separate folder and then start the slideshow (unless I'm missing something).

PR #2058, opened 2021-01-24
Apply selection to slideshow #1958

tanvir-ahmod:


tibbi:
I dont really want to have such slideshows of selected items, could be messy

PR #2096, opened 2021-02-28
Tamil language translation

GobinathAL:
I have included tamil language translation in strings.xml under values-ta


tibbi:
yeah looks good, thanks :)

PR #2105, opened 2021-03-16
remove okhttp dependency from runtime classpath to save appsize

connyduck:
OkHttp is pulled in by Picasso as a transitive dependency, but this app doesn't even have the internet permission, so it makes no sense to ship a whole networking library with the apk.

This change saves **133kB** in apk size for the floss-release flavor.

reference https://github.com/square/picasso/issues/2192

ps: Do you think it is possible to remove either Picasso or Glide completely from the app? Why was it necessary to have 2 image loaders in the first place?


tibbi:
one is used for backup, if the second one fails. I want to keep it like that, the file size doesnt really matter at all

PR #2106, opened 2021-03-16
enable R8 fullmode and cleanup shrinker rules

connyduck:
This change reduces apk size by **589kB** on the fossRelease variant. 
I used the app with these changes for a while with no problems. I might not have used all features though.

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -1,20 +1,3 @@
--keep class com.simplemobiletools.** { *; }
--dontwarn android.graphics.Canvas
--dontwarn com.simplemobiletools.**
--dontwarn org.apache.**
-
-# Picasso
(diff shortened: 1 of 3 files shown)


app/proguard-rules.pro
```diff
@@ -1,20 +1,3 @@
...
-
--keepclassmembers class * implements android.os.Parcelable {
-    static ** CREATOR;
-}
-
-# RenderScript
--keepclasseswithmembernames class * {
-native <methods>;
-}
--keep class androidx.renderscript.** { *; }
```
connyduck, line 20: t+These rules are not necessary because they are either included in `proguard-android-optimize.txt` or the libraries themself.

tibbi:
too dangerous, for example the photo editor broke right away

PR #2127, opened 2021-05-07
Icons in SettingsActivity

GobinathAL:
I am planning to add icons in SettingsActivity. Here is a screenshot. Can i proceed with it?
image


tibbi:
thanks, but I dont really want icons there

PR #2139, opened 2021-05-21
Update strings.xml

Yngvar-Skjaldulfsson:
Improved some translations to spanish so they sound more natural and not literally translated


tibbi:
nice, thanks

PR #2140, opened 2021-05-22
Fixed an issue in landscape mode

GobinathAL:
### Steps to reproduce:
* open the app
* change the device orientation to landscape
* click filter media icon in the menu bar
* Portrait option is not accessible for devices with low dpi


tibbi:
looks good, thanks

PR #2159, opened 2021-06-17
Avoid SQLite connection object leaking

DaPa:
- W/SQLiteConnectionPool: A SQLiteConnection object for database '/data/user/0/com.simplemobiletools.gallery.pro/databases/gallery.db' was leaked!  Please fix your application to end transactions in progress properly and to close the database when it is no longer needed.

Happens to me on Android 11 (Moto g 5G plus, with Google Photos disabled):
 - open Camera, then click on the small picture thumbnail from bottom-right corner which should open the Gallery app
 - in this scenario sometimes, the Simple Gallery was crashing and the above warning was seen in the logs...
 - no more crash after this patch: please review/test properly before incorporating!


tibbi:
I made it a bit simpler, guess it won't break anything. Thanks

PR #2171, opened 2021-07-03
Brazilian Portuguese translation updated

RodolfoCandido:
values-pt-rBR/strings.xml updated to latest (May 13, 2021) values/strings.xml


tibbi:
cool thanks

PR #2173, opened 2021-07-04
fix.xml

spkprs:


spkprs:
...were is the string "ADD TEXT" in editor?

tibbi:
not sure, maybe somewhere in the library. Ill see if there are any new strings available

PR #2175, opened 2021-07-05
Updated Hungarian language

hevesij:


tibbi:
use spaces instead of tabs and update only the strings that are changing, not all of them. It is hard to see what actually changed now. Instructions can be found at https://github.com/SimpleMobileTools/General-Discussion#how-can-i-suggest-an-edit-to-a-file . Thanks

Issue #2192, opened 2021-08-03
Bug: Video player buttons on the bottom work while they are hidden


efb4f5ff-1298-471a-8973-3d47447115dc:
https://user-images.githubusercontent.com/73130443/127983587-ce4c9f9f-e984-4a30-95af-6c682e1be553.mp4

PR #2197, opened 2021-08-09
Disable action buttons on the bottom while they are hidden. Ref #2192

serAndroidDev:
I fixed the issue, please check.


tibbi:
too lazy solution, just creates another glitch

PR #2198, opened 2021-08-17
Feature/lock swipe

jhnaiem:
Sometimes we give our phone to others to see an image, but they start to swipe. For this reason, added a new feature(Toggle Swiping)  so that users can lock and unlock swiping option.


tibbi:
Thanks, but I dont really want to add such button. And the code is written in a weird style too.

jhnaiem:
Dear ,
I am new to coding not that much aware of coding style . Can you please
suggest what should I do ? I respect you comment but if you can delete the
comment about coding style please .

Thank you .

PR #2208, opened 2021-08-23
Adding quick file overwrite option during saving edited images and mo…

aleksey-kuskov:
Adding quick file overwrite option during saving edited images and moving original file to recycle bin if configured


tibbi:
This is actually only for the old editor and it is really raw, would need too much work to make it work well, so I cannot accept it now. Thanks though

Issue #2236, opened 2021-10-07
Batch rotate without permissions fails with exception


jenda69:
Hello!
I just got SD card for my wife's phone and couple of hours later, she came to me with a problem.

If she chose one or multiple photos in a directory on the SD card and then used Menu-Rotate (in directory view), she just got multiple of toast popups with errors. Simple Gallery didn't have permission to write to the SD card.

After some fiddling around in the permissions menu, clearing data etc., I managed to solve this issue by opening a single photo (in full screen view) and tapping the Rotate button. Simple Gallery then offers a system dialog to gain required permissions to the SD card.

It probably should offer this dialog from a directory view too.

jenda69:
It's the latest version 6.21.3 Pro on Android 11 / MIUI 12.
But as I wrote, I rotated single photo and I got permissions like that and all is solved and well now.

I'm just reporting this because if you first rotate multiple images without permissions, you just get bunch of error and exceptions instead of permission dialog.

PR #2257, opened 2021-11-06
Fixed missing permission for rotate on external memory (#2236)

Aga-C:
Hi,

For every action that was saving anything on external memory, the app was checking permission and asking for it if needed. However, it wasn't done for batch rotating, so I've added it. It's done similarly as in other places. 

Some small notes about implementation:
- We are always rotating in the same directory, so it's enough to check permission for just one file.
- To not repeat the code between handleSAFDialog callback and no permissions required, I've moved the rotation code to a variable holding a function.
- I've moved "Saving" toast to be displayed just before saving, not at the start of the function.

**Before:**

https://user-images.githubusercontent.com/85929121/140605444-ba0b7eaf-ba94-4c98-9237-219d3d0a58b9.mp4

**After:**

https://user-images.githubusercontent.com/85929121/140605451-a5d1b573-2b8d-4fe7-9ec7-d79be5a3b6fc.mp4

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/gallery/pro/adapters/MediaAdapter.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/gallery/pro/adapters/MediaAdapter.kt
@@ -307,36 +307,37 @@ class MediaAdapter(
         }
     }
 
-    private fun rotateSelection(degrees: Int) {
-        val paths = getSelectedPaths().filter { it.isImageFast() }
+    private fun handleRotate(paths: List<String>, degrees: Int) {
         var fileCnt = paths.size
         rotatedImagePaths.clear()
-        val handleRotate = {
-            activity.toast(R.string.saving)
-            ensureBackgroundThread {
-                paths.forEach {
-                    rotatedImagePaths.add(it)
-                    activity.saveRotatedImageToFile(it, it, degrees, true) {
-                        fileCnt--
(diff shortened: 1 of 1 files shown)


app/src/main/kotlin/com/simplemobiletools/gallery/pro/adapters/MediaAdapter.kt
```diff
@@ -308,24 +308,36 @@ class MediaAdapter(
...
-                activity.saveRotatedImageToFile(it, it, degrees, true) {
-                    fileCnt--
-                    if (fileCnt == 0) {
-                        activity.runOnUiThread {
-                            listener?.refreshItems()
-                            finishActMode()
+        val paths = getSelectedPaths().filter { it.isImageFast() }
+        var fileCnt = paths.size
+        rotatedImagePaths.clear()
+        val handleRotate = {
```
tibbi, line 314: this looks so weird, better just make it a casual function and call it

tibbi:
ye should be good, thanks

tibbi:
if you write "fix #2236" at the commit message it will automatically close the issue on accepting the pull request, so use it please as I sometimes forget closing the proper issues

Aga-C:
Ok, I didn't know about that. I've just added comments to issues that I've done and that are still open. Please take a look at them and close :)

Issue #2102, opened 2021-03-14
Mark favorites in overview (star)


PinkPlasticBeach:
Hello
And first of all thanks for the many nice "simple" apps in orange.

I like the simplicity of this APP but Theres one thing im really missing:
When marking photos AS Favorites (star) i find them in a special Album/folder but i cant See its a favorite Ehen scrolling through the overview.
This would be great for sorting out photos when theres many that are alike and i just want to keep one/two best shots. I usually swipe through the photos in fullscreen mode and give the ones i really like a star. But going back to overview i cant identify the ones i chose, cause theyre so small i cant See the difference. I would like to select multiple photos in overview and delete all that look alike, keeping the favorites.

I believe thats not hard but would make the APP a lot better for me.
I think it should be default with opt-out in the settings.

Next nice thing would be writing the star to exif-info of the image to be able to quickly identify favorites if i dump the whole collection to my desktop...but thats another wish actually.

Thanks again for these nice tools, kindly
Plastic Beach

AlbatorV:
I reuest this too. Please add a litlle star icon over image when browsing folders.

PR #2258, opened 2021-11-07
Added stars to mark favorite items (#2102)

Aga-C:
Hi,

I've added marking favorite items with stars, so it's visible which items from a particular album are added to favorites. I've added it in both grid view and list view in the top-right corner. In case of grid view, I had to move video duration to the bottom-right corner of the thumbnail.

image


tibbi:
try adding more padding to the star, looks too big. I just added a new string mark_favorite_items, please add a new toggle into the app settings -> File thumbnail style for this pull request

tibbi:
guess it can be enabled by default, just test it properly with other thumbail items like file type, video duration, selection check etc, to make sure they dont overlap

Aga-C:
I've added a setting, made star smaller, and also moved everything on a thumbnail around to not overlap anything.

image

app/src/main/res/layout/photo_item_grid.xml
```diff
@@ -17,7 +17,7 @@
         android:layout_width="wrap_content"
         android:layout_height="wrap_content"
         android:layout_alignParentStart="true"
-        android:layout_alignParentTop="true"
+        android:layout_alignParentBottom="true"
```
tibbi, line 20: you cant just change something like this, it now fully overlaps with the file name

Aga-C:
Should I move everything above the file name if it's visible?

tibbi:
no, that would hide the image then. Keep icons at the top, not bottom. Lets move the video duration to the right of the Video icon and Favorite at the top right corner. Make the Video and Favorite icons the same size and margins though so they are symmetric.

Aga-C:
I've changed it.

image

tibbi:
ye should be good, Ill just remove the new dimens values, I want to avoid adding useles new values like that. Ill do that in master though, it is easier to edit multiple files like that. Thanks

PR #2259, opened 2021-11-12
Update strings.xml (Turkish)

ersen0:


tibbi:
nice, thanks for the updates :)

Issue #2281, opened 2021-12-04
Count files in headers


AlbatorV:
Hi
When you group files in folder, can you add files counter in parenthesis for each group headers

e.g. If group by month…

Fevrier 2021 (14)
Janvier 2021 (24)
Décembre 2020 (6)

Regards

Aga-C:
@tibbi What about adding this as an optional checkbox in the Group By dialog box?

![image](https://user-images.githubusercontent.com/85929121/144856632-bc80b473-9569-4d27-803a-669e1775044f.png)

AlbatorV:
+1
Yes, It's the good place for this option

tibbi:
ye, you can add it there as "Show file count at section headers"

PR #2283, opened 2021-12-04
Added file count in group headers (#2281)

Aga-C:
Hi,

I've added a simple file counter for group headers.

Screenshot_1


AlbatorV:
Waou excellent !!! 1h after my request !!!
And I see the new fav icons in thumbnail.

tibbi:
I dont want to add it by default, looks way too unnecessary. Maybe a setting in the future or so.

Issue #2333, opened 2022-01-13
Add shortcuts to long tap icon


AlbatorV:
It will be interresting to add shortcuts to long tap icon for Camera and Favorites, maybe recycle bin…

tibbi:
not sure what you mean, long tap where? You can already add home screen shortcuts on Android 8+.

AlbatorV:
Long tap in app icon on launcher/drawer
https://guides.codepath.com/android/Creating-App-Shortcuts-in-Launcher

PR #2335, opened 2022-01-14
Add support for App Shortcuts (#2333)

sidhantailawadi:
Added functionality for #2333

Screenshot_1642190374

Supports recycle bin & favorites for now

1) Shortcut icons require a darker color to be visible. Had to add two vectors . Should these be moved to commons?
2) If user has no favorites/recycled items, and user clicks on shortcut for favorite/recycle bin; it shows "no media found for filter" and instantly goes back to MainActivity.  How should this be ideally handled?
3) Need to support different variants for targetClass/targetPackage

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -301,7 +301,10 @@
             android:icon="@mipmap/ic_launcher_red"
             android:roundIcon="@mipmap/ic_launcher_red"
             android:targetActivity=".activities.SplashActivity">
-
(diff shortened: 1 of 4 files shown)


tibbi:
these should be created dynamically, no? Favorites shouldnt show up if there are no files marked as favorites, Recycle bin shouldnt show up if empty.

sidhantailawadi:
Yeah, makes sense. Too many issues in this PR.. Closing for now

PR #2357, opened 2022-02-15
fix: show delete animation after file delete

KryptKode:
**Notes**
- show delete animation after file delete

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/gallery/pro/adapters/MediaAdapter.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/gallery/pro/adapters/MediaAdapter.kt
@@ -459,27 +459,35 @@ class MediaAdapter(
             return
         }
 
-        val SAFPath = getSelectedPaths().firstOrNull { activity.needsStupidWritePermissions(it) } ?: getFirstSelectedItemPath() ?: return
+        val selectedItems = getSelectedItems()
+        val selectedPaths = selectedItems.map { it.path } as ArrayList<String>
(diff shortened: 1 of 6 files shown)


tibbi:
oh I somehow missed this PR, you havent updated the Trello card :)

tibbi:
as I wrote in Trello, let's not wait for the actual delete to be completed, just for the dialogs to be handled. It causes too much delay.

app/src/main/kotlin/com/simplemobiletools/gallery/pro/adapters/MediaAdapter.kt
```diff
@@ -475,10 +475,11 @@ class MediaAdapter(
             }
 
             media.removeAll(removeMedia)
-            listener?.tryDeleteFiles(fileDirItems)
+            listener?.tryDeleteFiles(fileDirItems){
+                removeSelectedItems(positions)
+                currentMediaHash = media.hashCode()
+            }
```
tibbi, line 481: we should check the new SDK 30 dialog above these lines at handleSAFDialog too, to not do anything before it is handled properly

KryptKode:
Reverted the changes before and now handling the SDK 30 dialog before the animation

tibbi:
works well in general, it just glitches a bit. If I delete a single file from the file thumbnails view for the first time and grant the SDK 30 permission, no delete animation is displayed properly. It just blinks a bit and the file that was supposed to be removed is still shown. I guess that the animation actually starts, but the files get refreshed too early, before it is actually deleted. If I do pull-to-refresh, the file properly disappears instantly after refreshing, without an animation. Can you reproduce it?

KryptKode:
Will fix that in another PR. 🙌🏾

PR #2379, opened 2022-03-13
SDK 30 changes

KryptKode:
**Notes**
- show an error when a user tries to pick the root directory of any volume
- the error shown is the string `copy_to_restricted_folder_message` which has the text `Unable to use this folder.\nThe system does not allow us to copy/move to this folder``
- handle SAF for SDK 30 before copying/moving files


tibbi:
I shortened the new strings as Android 12 seems to limit toasts to 2 lines max. Had to shorten one in the calendar too, we will apparently have the error toasts trimmed too.

PR #2386, opened 2022-03-23
Added "do not sleep" option in settings

DoozyDoz:
### Changes
- option in settings for the user to choose whether to keep on display when viewing media in full-screen mode

https://user-images.githubusercontent.com/98877504/159651439-b984ac28-93fc-4dea-b18d-11af08b29853.mp4


tibbi:
thanks, but that isnt really the way to go, and new strings should be added in all languages.

DoozyDoz:
Thanks for the work @tibbi 
I will translate the strings next time but what would be the better way to go about this if this is wrong?

tibbi:
well it certainly shouldnt be called Dont sleep. And you have wrong spacing at the code too, using empty lines randomly, not consistently within the app at all.

DoozyDoz:
Thank you for taking the time to reply, I will fix the issue and submit a new PR
Regards

Issue #2378, opened 2022-03-12
feature to keep the display on


sidmint:
An option to keep the display on while using the app. I usually read texts in iamges which takes longer than usual time. Thanks for the app.

PR #2387, opened 2022-03-24
[feat] Keep display on when viewing media in full screen mode 

DoozyDoz:
###Overview 
- In response to feature request #2378 and the changes requested in PR #2386.

###Changes
- removed unnecessary spaces 
- changed "do not sleep" to "keep display on"
- translated string to available languages.


tibbi:
add the strings into the proper place, not at the end of the files. And use just the english version, do not use machine translating.

DoozyDoz:
Thanks let me do so

DoozyDoz:
@tibbi - updated as requested.
Regards

PR #2402, opened 2022-04-10
Add padding into the grid view

zafarella:
* update to latest gradle
* fix minor gradle dep issue for room
* add padding in the directory and thumbnail view


tibbi:
you can already add padding properly in the app settings

zafarella:
ah, didn't noticed setting, ty. imho default should be 16

PR #2404, opened 2022-04-10
fix renaming

KryptKode:
### Changes
- show an error message when a user tries to rename a file on the root of the SD Card in `ViewPagerActivity` and `MediaAdapter`
- in `PickDirectoryDialog`, use the `Activity.isAStorageRootFolder` method instead of `String.isBasePath`

> The commons module should be updated after merging [the PR](https://github.com/SimpleMobileTools/Simple-Commons/pull/1369) for the changes in this PR to work

https://user-images.githubusercontent.com/25648077/162611064-fd92c21c-d729-4924-8f19-688cc5fb6c46.mp4

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/gallery/pro/adapters/MediaAdapter.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/gallery/pro/adapters/MediaAdapter.kt
@@ -234,11 +234,19 @@ class MediaAdapter(
     }
 
     private fun renameFile() {
+        val firstPath = getFirstSelectedItemPath() ?: return
+
+        val isSDOrOtgRootFolder = activity.isAStorageRootFolder(firstPath.getParentPath()) && !firstPath.startsWith(activity.internalStoragePath)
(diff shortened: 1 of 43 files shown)


app/src/main/kotlin/com/simplemobiletools/gallery/pro/adapters/MediaAdapter.kt
```diff
@@ -234,11 +234,19 @@ class MediaAdapter(
     }
 
     private fun renameFile() {
+        val firstPath = getFirstSelectedItemPath() ?: return
+
+        val isSDOrOtgRootFolder = activity.isAStorageRootFolder(firstPath.getParentPath()) && !firstPath.startsWith(activity.internalStoragePath)
+        if (isRPlus() && isSDOrOtgRootFolder) {
+            activity.toast(R.string.rename_in_sd_card_system_restriction)
```
tibbi, line 241: use Toast.LENGTH_LONG for such longer error message please, so there is enough time for reading them

PR #2405, opened 2022-04-11
restore from recycle bin SDK 30+

KryptKode:
### Changes
- when a file is on the root of Internal Storage or SD Card, change the destination to the `/Pictures` folder; we cannot write to the root of any volume on SDK 30+
- call `handleSAFDialogSdk30`, so users are prompted to allow SAF when restoring, if not previously allowed
- if there are conflicts, handle it as though it's a `CONFLICT_KEEP_BOTH` by appending some characters to the destination name
- add a new string `R.string.restore_to_path` shown when restoring to `/Pictures `folder on SDK 30+ instead of the root folder


tibbi:
tried restoring an existing file at /internal/DCIM/Camera/Subfolder, it created another one with (1) appended, didnt ask for conflict resolution. Lets keep it like this though, it isnt a bit issue and I dont want to postpone the new app update.

PR #2436, opened 2022-04-28
handle mange media for renaming

KryptKode:
## Notes
- handle manage media for renaming/deleting
- move the `launchMediaManagementIntent` method to the commons module
    - we added this in the commons module in [this PR](https://github.com/SimpleMobileTools/Simple-Commons/pull/1385) so we can detect when a user has activated or denied the permission to trigger a callback
    - we added a new method `BaseSimpleActivity.checkManageMediaOrHandleSAFDialogSdk30` that performs the check and determines if we should show the SAF dialog or not in [this PR](https://github.com/SimpleMobileTools/Simple-Commons/pull/1391)
 - the commons module should be updated after merging those PRs


tibbi:
it doesnt seem to work. At renaming with Media Management app on Android 12 I get a FileSystemException and at deleting I get the new dialog with "Use this folder"

tibbi:
tested with a folder inside /internal/DCIM/Camera

tibbi:
seems to work well in my tests, thanks

PR #2464, opened 2022-05-14
fix media store crash on deletion on android 12

### LINKED ISSUE #2454: Can't remove the only file in a folder from the image view
Hi,

I'm unable to remove the file from the image view in internal storage when it's the only file in its folder. 
- While removing, it throws an exception and then the only folder I'm able to see is a recycle bin (after refreshing the app, I can see all of them again). 
- However, removing still works when done from the thumbnails view. 
- Also, after unsuccessful remove, the file appears in recycle bin, but it's also in its original location.

https://user-images.githubusercontent.com/85929121/167111788-52bc5aea-7086-423c-8e46-307080452f9e.mp4

I have Android 12 and Simple Gallery 6.23.5 Pro (proprietary release from GitHub).

KryptKode:
## Notes
- MediaStore crashed on deletion because the `ViewPagerActivity` got finished before the `MediaStore.createDeleteRequest `executes completely
- To fix, we apply the same approach in the `MediaActivity` and finish explicitly after deletion completes, if the media list is empty
- also, only refresh the UI when the media list is not empty to avoid UI glitches
- fix the slideshow bug by resetting `mAreSlideShowMediaVisible` back to `false` when the slideshow ends
- closes #2454

**Before** | **After**
---|---
<video src="https://user-images.githubusercontent.com/25648077/168409270-fad5a29e-0ecd-4bf0-bb9f-6ea2efc7e41a.mp4" width="320"/> | <video src="https://user-images.githubusercontent.com/25648077/168409271-29c90842-5587-42fc-9b26-238d8dc636b0.mp4" width="320"/>

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/gallery/pro/activities/ViewPagerActivity.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/gallery/pro/activities/ViewPagerActivity.kt
@@ -591,6 +591,7 @@ class ViewPagerActivity : SimpleActivity(), ViewPager.OnPageChangeListener, View
             showSystemUI(true)
             mSlideshowHandler.removeCallbacksAndMessages(null)
             window.clearFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON)
+            mAreSlideShowMediaVisible = false
         }
     }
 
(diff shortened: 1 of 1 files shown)


tibbi:
looks good

Issue #2509, opened 2022-06-10
Viewing videos snaps back to first file in folder instead of desired file.


akeb1a:
Don't know when exactly this started happening, currently using 6.23.11 Pro, on Android ~~11~~ 10
Folder structure is like this:

SD card root
- Folder ABC
-- Folder 123
--- x number of video files
-- Folder 456
[etc]

I tap on any video file but the first, when the preview window opens, it displays the thumbnail of the chosen file for half a second before snapping back to display the first file in the folder (even if it's not a video file), and I have to manually swipe right all the way to whatever video I wanted to watch. This gets annoying quick in a folder with 30+ videos. I can prevent this by immediately playing the video before it snaps to the first file (or checking "Play videos automatically" in the settings), which seems to "interrupt" this behavior, allowing me to play the desired file, but doing so prevents me from being able to swipe right/left to the next or previous video afterwards, when you could do this in earlier versions.

This behavior happens in folders with one or more of the following types: .webm, .mp4, .mov files. 
Strangely, .mkv doesn't seem to be affected. 

edit: .gif files are also unaffected, so this should purely be something with video files.

hydrargyrum:
I experience the same problem for jpg files (I did not test videos): when selecting any image from a file manager, it briefly displays the chosen image (a few milliseconds) and quickly switches to the first image (by name) of the folder, which is very irritating because I won't swipe to reach the desired image when there are dozens (or hundreds!) of photos.

hydrargyrum:
Version 6.23.7 is the last version that works properly, all subsequent versions are buggy.

DrSaad77:
I am experiencing the same issue, from version 6.23.X !
You described the issue properly, I did not know at first, if was related to my files manager, or to Simple Galery!!
And each time, a new version is pushed and the issue persist.

DrSaad77:
> Don't know when exactly this started happening, currently using 6.23.11 Pro, on Android 11 Folder structure is like this:
> 
> SD card root
> 
> * Folder ABC
>   -- Folder 123
>   --- x number of video files
>   -- Folder 456
>   [etc]
> 
[...]

Issue #2508, opened 2022-06-10
Deleting a file while using an RTL language takes me back to first picture in folder


mhmdanas:
Steps to reproduce:
1. Switch to an RTL language (Arabic in my case)
2. Open any folder in Simple Gallery
3. Tap on an image to view it in fullscreen
4. Press on the trash icon at the bottom to delete the image, and confirm the deletion when the confirmation dialog is shown 

## Expected behavior

The picture next to the one just deleted in the folder is displayed (at least this is what happens when I set the system language to English).

## Actual behavior

The first picture in the folder is displayed.

DrSaad77:
It happens also to me, lets say there are 100 pic is a folder, and during sliding pic with SimpleGalery, I decided I dont like picture 34, and I want to delete it. After deleting it, hope I brings back and show the first picture of the folder!

Naveen3Singh:
Seems related to this line: https://github.com/duolingo/rtl-viewpager/blob/940f12724ff0fe624f12e8218b2b967891aa5396/src/main/java/com/duolingo/open/rtlviewpager/RtlViewPager.java#L98

maybe it's time to switch to ViewPager2

tibbi:
we can switch to whatever works

Issue #2332, opened 2022-01-11
Performance regression when opening from external file manager


tp0:
When opening an image from a file manager app ("open with" way), it takes about 2 seconds before the image is actually visible. Simple Gallery opens instantly but for those nearly 2 deconds displays the empty background. ([Screenshot](https://i.imgur.com/PLUIxRc.png))

* File is stored on internal storage. File size makes no difference, happens with very small files too.

* Issue occurs in 6.22.3 and 6.22.2, but not in 6.22.1 or 6.22.0.

* File managers tested: Simple File Manager Pro (6.11.3), Material Files

Device: Motorola G8 Power
Android: 11

tibbi:
the gallery fetches all files in the given folder to make it swipable, not just the given 1 image. I guess if you use Simple File Manager you can swipe to other images too, but with Material files you cannot, right? Or how does it work, is it slow in both cases?

PR #2522, opened 2022-06-18
Fix position snap glitch

Naveen3Singh:
Closes https://github.com/SimpleMobileTools/Simple-Gallery/issues/2509, https://github.com/SimpleMobileTools/Simple-Gallery/issues/2508

Tested with hidden folders, renamed  & new folders, and empty database caches and the bug seems fixed....
... but it's not confirmed by any bug reporters yet. 
Also tested the delete, and move operations as similar position snapping bugs were reported and fixed before.

This PR ~~won't fix~~ also fixes the position snapping bug when using an RTL language as mentioned in [this](https://github.com/SimpleMobileTools/Simple-Gallery/issues/2508) issue

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/gallery/pro/activities/ViewPagerActivity.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/gallery/pro/activities/ViewPagerActivity.kt
@@ -59,6 +59,7 @@ import kotlinx.android.synthetic.main.bottom_actions.*
 import java.io.File
 import java.io.OutputStream
 
+@Suppress("UNCHECKED_CAST")
 class ViewPagerActivity : SimpleActivity(), ViewPager.OnPageChangeListener, ViewPagerFragment.FragmentListener {
     private val REQUEST_VIEW_VIDEO = 1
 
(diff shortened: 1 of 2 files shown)


app/src/main/kotlin/com/simplemobiletools/gallery/pro/activities/ViewPagerActivity.kt
```diff
@@ -388,7 +371,17 @@ class ViewPagerActivity : SimpleActivity(), ViewPager.OnPageChangeListener, View
             }
         }
 
-        refreshViewPager()
+        // show the selected image asap, while loading the rest in the background to allow swiping between them. Might be needed at third party intents
+        if (mMediaFiles.isEmpty() && mPath.isNotEmpty() && mDirectory != FAVORITES) {
```
tibbi, line 375: you basically just removed the cache file displaying, while leaving the no longer valid comment there... That code was there for a reason, now it can take a long time for anything to be shown.

Naveen3Singh: line 374 is a valid comment, read lines 374 to 384. `refreshViewPager()` does the background loading instead.

The file refresh done in `refreshViewPager()` almost always executes the callback a little before `getCachedMedia()` does the db lookup. Confirmed with 100s of files a couple of times on a couple of devices. Sometimes the path specified in the external intent might not be cached in the db and `getCachedMedia()` might return empty or incomplete data and cause the viewpager to snap to some invalid position

The `getCachedMedia()` was introduced in PR https://github.com/SimpleMobileTools/Simple-Gallery/commit/318664fca1b751dc6f14140b3d0c6cf91fa934f7 for external intents as a fix for https://github.com/SimpleMobileTools/Simple-Gallery/issues/2332 (files were shown after the loading was complete thus delayed)...

> now it can take a long time for anything to be shown

... so that still works as per my testing and as per lines 374-382.

tibbi: well ok, that hackfix has been added for the previous way of fetching files, lets see what will the user report back at th given issue

Naveen3Singh:
It turns out changing the adapter caused the `currentItem` to reset to zero and since the page change listener is active, that resets the `mPos` value to zero, and `setCurrentItem()` is set to 0 again.

https://github.com/SimpleMobileTools/Simple-Gallery/pull/2522/commits/e6039d07cb7d65ad370b6b2d416c482972474e80 fixes this.

due to some internal viewpager voodoo this happens in RTL layouts only, either way it's a good idea to remove the listener before changing the adapter

tibbi:
alright thanks, lets see how will it work in the real world :)

PR #2543, opened 2022-07-05
Properly handle media loading and cache 

Naveen3Singh:
### Changes:

- Clear media cache only for external intents, previously it was cleared in all cases.
- When the intent is external:
  - Do sort the items randomly when random sorting is enabled
  - Allow refresh even when the video is playing (for devices where the media is loaded with delay or when there are too many items to load)

No side effects so far.


tibbi:
have you tested it with third party app intents too? Like launching a video from some file manager?

Naveen3Singh:
Yes, tested with two file managers on three devices. Didn't find any issues with delayed loading or random sorting

tibbi:
random sorting is too random now, it shows the files in a different order at the MediaActivity and ViewPagerActivity

tibbi:
Like if you open even an image from MediaActivity and swipe to the sides, it doesnt respect the shown file thumbnail order

Naveen3Singh:
Can you pls attach Prefs.xml?

tibbi:
it is used only at sorting by path or name, it makes "file2" appear before "file100".

Naveen3Singh:
it was my bad, an inverted boolean. fixed in e4aff60

app/src/main/kotlin/com/simplemobiletools/gallery/pro/activities/ViewPagerActivity.kt
```diff
@@ -1228,7 +1229,8 @@ class ViewPagerActivity : SimpleActivity(), ViewPager.OnPageChangeListener, View
     }
 
     private fun refreshViewPager(refetchPosition: Boolean = false) {
-        if (config.getFolderSorting(mDirectory) and SORT_BY_RANDOM == 0) {
+        val isNotRandomSorting = config.getFolderSorting(mDirectory) and SORT_BY_RANDOM == 0
+        if (isNotRandomSorting || isExternalIntent()) {
```
tibbi, line 1233: please make this !isRandomSorting, it looks weird

Naveen3Singh: [all right :)](https://github.com/SimpleMobileTools/Simple-Gallery/pull/2543/commits/caf6b7c5535f653ad10b997526758c6226d6a505)

tibbi:
I mean that reordering files at clicking on a folder thumbnail is ok, but at clicking a file thumbnail they shouldnt change their order

Naveen3Singh:
I reinstalled to test it on two of my devices but could not reproduce this after the changes done in https://github.com/SimpleMobileTools/Simple-Gallery/commit/e4aff60852d9564360b0a1f3be975bd486ed5b2c

Are you sure you have the build with the latest commit https://github.com/SimpleMobileTools/Simple-Gallery/commit/e4aff60852d9564360b0a1f3be975bd486ed5b2c or https://github.com/SimpleMobileTools/Simple-Gallery/pull/2543/commits/caf6b7c5535f653ad10b997526758c6226d6a505?

(for third-party intents, there is no media cache (as there is no media activity) and content is reshuffled in that case)

tibbi:
yeah, using the latest version on Android 11, but I dont think it matters. Test it with images, not videos. And use more files, not just 3.

tibbi:
it happens with fresh install, so everything is default, just set Random file sorting

Naveen3Singh:
still testing but seems fixed :)

tibbi:
alright, should be good, will try releasing it :) Thanks

Issue #2438, opened 2022-04-30
[improvement] option to replace current file when editing (checkbox?) or setting to set default file name as current file


tcaERROR-issues:
99% of edits I do on a cellular device is cropping screenshots of which I have no need for the original.

I would like to see a streamlined way to save&replace which doesnt require me to load my keyboard and backspace "_1"

Thanks for a brillaint suite of apps.

otede:
+1
On that note I really appreciate the fact that cropping no longer touches the file date!

PR #2546, opened 2022-07-06
[feat] - Replace original file after edit

DoozyDoz:
### Description
- Address feature request in issue #2438 
### Screenshot
Screenshot_20220503-155649

### Changes
- added a checkbox to the "save as" dialog that a user can tick in case the user wants to replace the original image when saving.
- upon ticking the checkbox the filename and extension fields are disabled.
- The dialog positive button is changed from "OK" to "Replace".
- When a user clicks "Replace", they are not prompted with the "overwrite confirmation" dialog (for the process of replacing to happen in fewer clicks, I imagine the user that selects "Replace Original File" option is already sure of what they want)


DoozyDoz:
@tibbi please revert

DoozyDoz:
> revert what?

If this fixes the issue.

tibbi:
closing this for now, will consider some fix later. Thanks though

DoozyDoz:
Alright

PR #2552, opened 2022-07-12
Add autodelete of exif

dri94:
Auto-delete Exif data from every image in the gallery and new images added to the gallery. For the most privacy-conscious amongst us.


tibbi:
I dont really want to add such features, thanks though

dri94:
@tibbi Any particular reason? it's very little to maintain while having a huge boost for privacy by default.

PR #2569, opened 2022-07-24
Fix some deprecations

TacoTheDank:
I fixed some deprecations, including some with ExoPlayer. All of these are one-for-one replacements.


tibbi:
some will be replaced soon, thanks

PR #2615, opened 2022-09-20
fix wrong rotation for horizontally flipped images

KryptKode:
# Why
Horizontally flipped images with orientation [`ORIENTATION_TRANSVERSE`](https://developer.android.com/reference/android/media/ExifInterface#ORIENTATION_TRANSVERSE) were rotated by 0 degs in the ViewPager screen. 
Ideally, they should be rotated by 90. 

This PR fixes it.


tibbi:
thanks, added ORIENTATION_TRANSPOSE too, hope it wont break anything

Issue #2605, opened 2022-09-08
Slideshow "random" order displays the wrong filenames


aksrikanth:
If you select the slideshow on a folder with the "random order" checked, the images are shown correctly in random order, but the filenames are incorrectly shown in the default sort order.

Any file operations (copy, move, rename etc) will affect not the image displayed, but the filename that is shown. This is very unexpected.

Steps to reproduce:

1. Select folder with images
2. Select menu -> Slideshow
3. Ensure "Random order" checkbox is checked
4. Select OK
5. Tap image to see filename

Expected result: Random filename

Actual result: Filenames in default sort order

tibbi:
what app and OS version are you using?

aksrikanth:
OS: Android 13
App version 6.25.0 Pro (from f-droid)

Naveen3Singh:
@aksrikanth if you are up for testing, here's an test APK with fix: https://github.com/Naveen3Singh/Simple-Gallery/releases/download/slideshow_test/gallery-375-foss-debug.apk

aksrikanth:
This bug seems to be fixed in the debug build.

PR #2616, opened 2022-09-22
Reset sort order after stopping random slideshow

Naveen3Singh:
Closes https://github.com/SimpleMobileTools/Simple-Gallery/issues/2605


Naveen3Singh:
Resetting sort order got too messy. https://github.com/SimpleMobileTools/Simple-Gallery/pull/2617 is the correct and simpler solution for this.

PR #2628, opened 2022-10-11
Fix video partly off-screen in horizontal, RTL bug

Amir-yazdanmanesh:


tibbi:
so you tested if it works, right?

Amir-yazdanmanesh:
Yes I tested , it works. you can set language of your phone to arabic or persian and test again

Issue #2580, opened 2022-08-08
can allway chose to "search all visible folders" 


jorismq:
can make this to an option?
 thx

PR #2629, opened 2022-10-15
 [feat] Add setting to enable searching of all files by default (#2580) 

sidhantailawadi:
Implement the feature requested in #2580 

Added the setting for this under "File Operations"
Screenshot_1665847086


tibbi:
added a new string show_all_folders for it, but also make sure the value is properly exported and imported

tibbi:
actually I implemented it already

PR #2651, opened 2022-10-28
fix grammatical mistake in english strings

kmein:


tibbi:
thats fine

kmein:
I'm afraid it's not. It's either missing an apostrophe (all folders' content = the content _of_ all folders [genitive]) or the -s has to be moved around.

PR #2696, opened 2022-12-25
Changed show hidden button in folder picker to include excluded (#2693)

### LINKED ISSUE #2693: No Way to View Excluded Folders When Moving Files
Since you have formally added a button under the 3-dot menu to temporarily show excluded folders (something [I requested](https://github.com/SimpleMobileTools/Simple-Gallery/issues/2432#issuecomment-1125469375) back when the decision to remove hidden files was discussed, and you [announced](https://github.com/SimpleMobileTools/Simple-Gallery/issues/2503#issuecomment-1146892767) would be added, which made me very happy), I have noticed that the "Eye" button shown when moving files only reveals hidden folders, but there is no way to similarly reveal excluded folders when moving or copying files.
Perhaps, they could be combined into the same button?

Aga-C:
Closes #2693

**Before:**

https://user-images.githubusercontent.com/85929121/209464611-167777bd-5512-476b-81ee-1988f6b452ad.mp4

**After:**

https://user-images.githubusercontent.com/85929121/209464625-e27d3cb6-82a6-4c9a-9394-c46dd68a72a8.mp4

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/gallery/pro/dialogs/PickDirectoryDialog.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/gallery/pro/dialogs/PickDirectoryDialog.kt
@@ -72,8 +72,8 @@ class PickDirectoryDialog(
         fetchDirectories(false)
     }
 
-    private fun fetchDirectories(forceShowHidden: Boolean) {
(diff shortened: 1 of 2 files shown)


tibbi:
looks good, thanks.

PR #2848, opened 2023-05-23
Add strings related to resizing multiple images

Naveen3Singh:
`Enter a number between 10 and 90`: Used when the user enters a number outside the 10-90 range.
`Failed to resize %d images`: Used to show a toast when resizing some or all images failed.


tibbi:
would be better to have `failed_to_resize_images` as a plural form in case it is 1 image only, or multiple

Naveen3Singh:
Please check the Slovak singular translation again

tibbi:
its ok :) Thanks

PR #2851, opened 2023-05-24
Add option to bulk resize images

Naveen3Singh:
Users can utilize this feature to resize images in bulk, doesn't matter how many. The option to bulk resize images is visible when the following conditions are met:

1. At least one image is selected (SVGs and GIFs do not qualify)
2. The parent directory of the selected images is not a restricted directory. For example, in Android 11, the Download folder is restricted. Alternatively, if the app has the "Manage all files" permission, the resize option will still be available even in restricted directories.
3. All selected images share a common parent directory. This means that the files were not selected while in "Show all folders" mode. However, if the app has the "Manage all files" permission, the resize option will be accessible regardless of whether the selected images have a common parent directory.

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/gallery/pro/adapters/MediaAdapter.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/gallery/pro/adapters/MediaAdapter.kt
@@ -144,6 +144,7 @@ class MediaAdapter(
             findItem(R.id.cab_open_with).isVisible = isOneItemSelected
             findItem(R.id.cab_edit).isVisible = isOneItemSelected
             findItem(R.id.cab_set_as).isVisible = isOneItemSelected
+            findItem(R.id.cab_resize).isVisible = selectedItems.all { it.isImage() }
             findItem(R.id.cab_confirm_selection).isVisible = isAGetIntent && allowMultiplePicks && selectedKeys.isNotEmpty()
             findItem(R.id.cab_restore_recycle_bin_files).isVisible = selectedPaths.all { it.startsWith(activity.recycleBinPath) }
             findItem(R.id.cab_create_shortcut).isVisible = isOreoPlus() && isOneItemSelected
@@ -179,6 +180,7 @@ class MediaAdapter(
             R.id.cab_open_with -> openPath()
             R.id.cab_fix_date_taken -> fixDateTaken()
             R.id.cab_set_as -> setAs()
+            R.id.cab_resize -> resize()
             R.id.cab_delete -> checkDeleteConfirmation()
         }
     }
@@ -284,6 +286,20 @@ class MediaAdapter(
     private fun setAs() {
         val path = getFirstSelectedItemPath() ?: return
         activity.setAs(path)
+    }
+
+    private fun resize() {
+        val paths = getSelectedPaths()
+        if (isOneItemSelected()) {
+            val path = paths.first()
+            activity.launchResizeImageDialog(path) {
+                finishActMode()
+            }
+        } else {
+            activity.launchResizeMultipleImagesDialog(paths) {
+                finishActMode()
+            }
+        }
     }
 
     private fun toggleFileVisibility(hide: Boolean) {
--- a/app/src/main/kotlin/com/simplemobiletools/gallery/pro/adapters/MediaAdapter.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/gallery/pro/adapters/MediaAdapter.kt
@@ -30,14 +30,9 @@ import com.simplemobiletools.gallery.pro.interfaces.MediaOperationsListener
 import com.simplemobiletools.gallery.pro.models.Medium
 import com.simplemobiletools.gallery.pro.models.ThumbnailItem
 import com.simplemobiletools.gallery.pro.models.ThumbnailSection
-import kotlinx.android.synthetic.main.photo_item_grid.view.*
-import kotlinx.android.synthetic.main.thumbnail_section.view.*
+import kotlinx.android.synthetic.main.photo_item_grid.view.file_type
+import kotlinx.android.synthetic.main.thumbnail_section.view.thumbnail_section
 import kotlinx.android.synthetic.main.video_item_grid.view.*
-import kotlinx.android.synthetic.main.video_item_grid.view.favorite
-import kotlinx.android.synthetic.main.video_item_grid.view.media_item_holder
-import kotlinx.android.synthetic.main.video_item_grid.view.medium_check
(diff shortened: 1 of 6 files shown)


app/src/main/kotlin/com/simplemobiletools/gallery/pro/dialogs/ResizeMultipleImagesDialog.kt
```diff
@@ -0,0 +1,118 @@
...
+                    positiveButton.setOnClickListener {
+                        val resizeFactorText = resizeFactorEditText.text?.toString()
+                        if (resizeFactorText.isNullOrEmpty() || resizeFactorText.toInt() !in 10..90) {
+                            activity.toast(R.string.resize_factor_error)
+                            return@setOnClickListener
+                        }
+
+                        val resizeFactor = resizeFactorText.toFloat().div(100)
+
+                        alertDialog.setCanceledOnTouchOutside(false)
```
tibbi, line 57: you can define this as a setupDialogStuff parameter

app/src/main/res/layout/dialog_resize_multiple_images.xml
```diff
@@ -0,0 +1,41 @@
...
+
+    <com.simplemobiletools.commons.views.MyTextInputLayout
+        android:id="@+id/resize_factor_input_layout"
+        android:layout_width="match_parent"
+        android:layout_height="wrap_content"
+        android:layout_marginHorizontal="@dimen/big_margin"
+        android:layout_marginTop="@dimen/normal_margin"
+        android:hint="@string/resize_factor"
+        app:errorEnabled="true"
+        app:helperText="Resize images to the given percentage, value must be within 10 and 90"
```
tibbi, line 25: lets use the string resource here

app/src/main/kotlin/com/simplemobiletools/gallery/pro/adapters/MediaAdapter.kt
```diff
@@ -286,6 +283,32 @@ class MediaAdapter(
         activity.setAs(path)
     }
 
+    private fun resize() {
+        val paths = getSelectedItems().filter { it.isImage() }.map { it.path }
```
tibbi, line 287: the isImage() check might not be enough as it includes webm and heic too, which we cannot resize. Not sure about other formats like bmp etc.

app/src/main/kotlin/com/simplemobiletools/gallery/pro/dialogs/ResizeMultipleImagesDialog.kt
Naveen3Singh: it's only set when the actual resize is in progress, the dialog is cancellable otherwise and allows the user to go back, and change their selection, or do something else.

app/src/main/res/layout/dialog_resize_multiple_images.xml
Naveen3Singh: right, I forgot.

app/src/main/kotlin/com/simplemobiletools/gallery/pro/dialogs/ResizeMultipleImagesDialog.kt
tibbi: okay

app/src/main/kotlin/com/simplemobiletools/gallery/pro/adapters/MediaAdapter.kt
Naveen3Singh: By webm, you meant webp right?  It's working with the following formats:
```kotlin
val photoExtensions = arrayOf(".jpg", ".png", ".jpeg", ".bmp", ".webp", ".heic", ".heif", ".apng", ".avif")
```

The array above is used by the isImageFast() method which in turn determines the `isImage()` result.

tibbi: ye, webp

Naveen3Singh: But heic and avif aren't loaded at all on emulators i.e. x86. On real physical devices, they are visible and can be resized.

tibbi:
seems to work well in the initial test, can we try keeping the files "last modified" value, if the setting "Keep old last modified value at file operations" is enabled?

tibbi:
actually it doesnt work at single file resizing either, not sure if it is a device glitch or it isnt implemented at all?

tibbi:
testing it on samsung with Android 13, All Files access granted

Naveen3Singh:
> actually it doesnt work at single file resizing either, not sure if it is a device glitch or it isnt implemented at all?

the Resize menu is visible, right? What exactly happens when you tap it?

tibbi:
> > actually it doesnt work at single file resizing either, not sure if it is a device glitch or it isnt implemented at all?
> 
> the Resize menu is visible, right? What exactly happens when you tap it?

I mean that the files Last Modified value changes, even though it shouldnt. Resize itself works just fine.

Naveen3Singh:
It is implemented just as before:

```kotlin
rescanPaths(paths) {
        fixDateTaken(paths, false)
        for (path in paths) {
            val file = File(path)
            val lastModified = file.lastModified()
            if (config.keepLastModified && lastModified != 0L) {
                File(file.absolutePath).setLastModified(lastModified)
                updateLastModified(file.absolutePath, lastModified)
            }
        }
    }
```

I think `file.lastModified()` returns the updated lastModified here because the file was already overwritten by the time this code is called. I'm working on a solution now.

Naveen3Singh:
Scanning the paths _after_ updating the last modified time (https://github.com/SimpleMobileTools/Simple-Gallery/pull/2851/commits/761ebee77bc926cde82283b72b036bdfd1075ba8) fixes the issue for O+ versions. Not sure why `File.setLastModified()` returns `false` on the older version, should I keep digging?

tibbi:
no sure what you mean it is broken for me on Android 13, not just some low versions

Naveen3Singh:
I mean I fixed the last modified time issue on newer versions (Oreo+)

tibbi:
oh ok, it works just fine. Sometimes after resizing I still see the old bigger resolution image at both thumbnail and fullscreen view, but that seems to be caused by the cache. It shows the proper images after clearing the app cache. We use the file name, last modified and size at getting the file signature for cache, so not sure why it misbehaves as it includes the size too. Anyway, it is good as it is for now, if it causes issues, we will get back to it. Thanks

PR #2859, opened 2023-05-29
Update.xml

spkprs:


tibbi:
seems like someone published translations via weblate already, update whatever is needed separately please

PR #2876, opened 2023-06-17
allow skipping recycle bin at deletion

fatihergin:
Allows quickly skipping the recycle bin on deletion confirmation dialog if user wants to delete files immediately.

It works with “Do not ask again in this session” option. 

If “Do not ask again in this session” is selected,  it's remembered the selected recycle bin skipping check and apply it. Either always skips the recycle bin, or don't.

If “Do not ask again in this session” option is not selected but new skipping option selected, only current action will be affected from selection. Neither the rest of the current session nor the next app sessions will not be affected from this selection.


saefthe:
نعم قم بلتكيد لا تسئلني مره اخرى

app/src/main/kotlin/com/simplemobiletools/gallery/pro/adapters/MediaAdapter.kt
```diff
@@ -505,15 +505,23 @@ class MediaAdapter(
...
         val isRecycleBin = firstPath.startsWith(activity.recycleBinPath)
-        val baseString = if (config.useRecycleBin && !isRecycleBin) R.string.move_to_recycle_bin_confirmation else R.string.deletion_confirmation
+        val baseString =
+            if (config.useRecycleBin && !config.tempSkipRecycleBin && !isRecycleBin) R.string.move_to_recycle_bin_confirmation else R.string.deletion_confirmation
         val question = String.format(resources.getString(baseString), itemsAndSize)
-        DeleteWithRememberDialog(activity, question) {
-            config.tempSkipDeleteConfirmation = it
-            deleteFiles()
+
+        val callback = fun(remember: Boolean, skipRecycleBin: Boolean) {
```
tibbi, line 512: lets add the function itself as a callback to DeleteWithRememberDialog as used in multiple other places, this is hard to read

fatihergin: like this?

```kotlin
val callback = object: DeleteWithRememberDialog.Callback {
    override fun onConfirm(remember: Boolean, skipRecycleBin: Boolean) {
        config.tempSkipDeleteConfirmation = remember
        ...
    }
}

DeleteWithRememberDialog(activity, question, config.useRecycleBin, callback)
```

tibbi: no, like
DeleteWithRememberDialog(activity, question, config.useRecycleBin) { rememer, skipRecycleBin ->
...
}
it is used in many dialogs, try finding one

tibbi:
yeah nice thanks, seems to work just fine

PR #2879, opened 2023-06-18
Feature/update exoplayer v2.18.7

fatihergin:


tibbi:
ok thanks, should be good

PR #2881, opened 2023-06-19
add search bar to folder picker at copy destination

fatihergin:


app/src/main/kotlin/com/simplemobiletools/gallery/pro/dialogs/PickDirectoryDialog.kt
```diff
@@ -32,6 +34,9 @@ class PickDirectoryDialog(
     private var isGridViewType = activity.config.viewTypeFolders == VIEW_TYPE_GRID
     private var showHidden = activity.config.shouldShowHidden
     private var currentPathPrefix = ""
+    private val config = view.context.config
+    private val searchView = view.folder_search_view
+    private val searchEditText = view.findViewById<EditText>(R.id.top_toolbar_search)
```
fatihergin, line 39: `view` is  an instance of `MySearchMenu` which is placed at `Simple-Commons`.
not a good practice to access an internal of custom view.

app/src/main/kotlin/com/simplemobiletools/gallery/pro/dialogs/PickDirectoryDialog.kt
```diff
@@ -72,6 +79,48 @@ class PickDirectoryDialog(
         fetchDirectories(false)
     }
 
+    private fun configureFolderSearchView() = with(searchView) {
+        updateHintText(context.getString(R.string.search_folders))
+        searchEditText.imeOptions = EditorInfo.IME_ACTION_DONE
```
fatihergin, line 84: can be moved to layout of `MySearchMenu`.

app/src/main/kotlin/com/simplemobiletools/gallery/pro/dialogs/PickDirectoryDialog.kt
```diff
@@ -72,6 +79,48 @@ class PickDirectoryDialog(
...
+
+        toggleHideOnScroll(!config.scrollHorizontally)
+        setupMenu()
+        onSearchClosedListener = {
+            searchEditText.clearFocus()
+            activity.hideKeyboard(searchEditText)
```
fatihergin, line 90: these two is already implemented in `MySearchMenu`, but not working properly if the view is used in an alert dialog. 
the fix can be applied on `Simple-Common`

PR #2882, opened 2023-06-21
remember last used copy destination on "Select destination" ->  "Other folder"

fatihergin:


app/src/main/kotlin/com/simplemobiletools/gallery/pro/dialogs/PickDirectoryDialog.kt
```diff
@@ -178,6 +180,24 @@ class PickDirectoryDialog(
         }
     }
 
+    private fun getOtherFolderOpeningPath(): String {
+        val lastCopyPath = config.lastCopyPath
+
+        val lastCopyPathExist = {
```
tibbi, line 186: do we really need this as a function? Looks quite weird

app/src/main/kotlin/com/simplemobiletools/gallery/pro/helpers/Config.kt
```diff
@@ -363,6 +363,10 @@ class Config(context: Context) : BaseConfig(context) {
         get() = prefs.getString(LAST_FILEPICKER_PATH, "")!!
         set(lastFilepickerPath) = prefs.edit().putString(LAST_FILEPICKER_PATH, lastFilepickerPath).apply()
 
+    var lastCopyPath: String
```
tibbi, line 366: after accepting lets move this, the constant and getOtherFolderOpeningPath to commons, so that we dont have to copypaste it into the file manager

app/src/main/kotlin/com/simplemobiletools/gallery/pro/dialogs/PickDirectoryDialog.kt
fatihergin: not really. just to make it consistent with `lastCopyPathVisible`. 
now it's updated by removing the function definition.

tibbi: and why is lastCopyPathVisible a function at all too? :) I meant both

PR #2886, opened 2023-07-03
Fixed exception on image resizing (#2883)

### LINKED ISSUE #2883: Image resizing bug
There is a bug in 6.26.9 and 6.26.10:
- resize image
- rename it before saving 
- having a bug message and thus, no saving new image!
P.S. When we don't rename resized image - its OK, and have no bug!
I have back in 6.26.8 and don't have this issue!

Aga-C:
Fixes #2883

Based on the old code from https://github.com/SimpleMobileTools/Simple-Gallery/pull/2851/files#diff-b7924de9715452539ac44202c12732810c03604d47f0e2ccdd8fc63465841533. In case of changed file name, the function was trying to read the bitmap from the new path instead of the old path, so it was needed to send both paths to `resizeImage`.


tibbi:
nice thanks

PR #2887, opened 2023-07-04
replace outdated image cropper library with an up to date fork

fatihergin:


tibbi:
seems to work just fine, thanks

PR #2898, opened 2023-07-21
Make subfolder grouping follow file structure

### LINKED ISSUE #1886: Subfolder Grouping Issue
Hello not sure if I will get help here but will try.
I purchased the Simply Gallery Pro and on the "Change View Type" option I checked the "Group direct Subfolders"
For some folders it seems to work and some folders don't.
For example - 
Folder A has Folder 1, 2, 3
Folder B has Folder X, Y, Z

Main screen shows
Folder A (3)
Folder X 
Folder Y
Folder Z

Shouldn't it be
Folder A (3)
Folder B (3) ??

Also Folders inside Folder B can't be moved get "Unknown error"
Any solution for this problem?

esensar:
This changes subfolder grouping to follow file hierarchy to prevent unreliable behavior with middle folders missing media.

This fixes #1886

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/gallery/pro/extensions/Context.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/gallery/pro/extensions/Context.kt
@@ -23,7 +23,6 @@ import com.bumptech.glide.load.resource.bitmap.RoundedCorners
 import com.bumptech.glide.load.resource.drawable.DrawableTransitionOptions
 import com.bumptech.glide.request.RequestListener
 import com.bumptech.glide.request.RequestOptions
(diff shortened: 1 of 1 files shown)


tibbi:
ok thanks, should be good

ujujiensan:
The original OP for #1886. 
So is there an update on the app? Don't see one yet.

tibbi:
theres no app update yet, the fix has been added just yesterday...

ujujiensan:
Ahh gotca. Thought there was an update for the fix. I wasn't sure how the process goes~
Can't  wait for the update and see if it is finally fixed.

Thank you.

PR #2900, opened 2023-07-21
Disable hide/unhide options for filesystem hidden folders

esensar:


esensar:
This disables hide/unhide options for folders that are hidden in filesystem (due to having name starting with `.`), because the way hiding works in this application will not change that fact, only renaming would. Not sure if we need to add some additional information to make it clearer to the users why that option is not present.

tibbi:
how are filesystems specified there? Wont those menu items be always visible?

esensar:
This PR will disable both hide and unhide options for folders that are hidden by having their name start with a dot, regardless of whether they have `.nomedia` file.

tibbi:
Im not sure if this is the correct way of doing it :/ Shouldnt we check if the folder starts with /android/data or so?

esensar:
Hmm, I am not sure if we are seeing the same problem then. The only folders for which I have seen confusing behavior are the ones whose name starts with a dot, meaning they are hidden (an example of such folder is WhatsApp Statuses - `.Statuses`). If you have example of application that behave like that without having `.` in their name, let me know so I can test that out too.

tibbi:
Im afraid this will make people write bad reviews or emails about Un/hide button missing, it needs some explanation. Having the hide button sometimes visible and sometimes not just isnt user friendly enough. How about changing it a bit and at pressing Unhide showing a basic confirmation dialog with text `If a folder or one of its parent folders has a dot before its name, it is hidden and cannot be unhidden like this. You have to remove the dot by renaming it.`. After the user confirms, just close the dialog, do not do anything. And do it at Unhide only, do not affect hiding functionality with this PR. Dont show anything, dont change when is the button visible.

app/src/main/kotlin/com/simplemobiletools/gallery/pro/adapters/DirectoryAdapter.kt
```diff
@@ -305,6 +305,11 @@ class DirectoryAdapter(
                 }
             }
         } else {
+            if (selectedPaths.any { File(it).isHidden }) {
```
tibbi, line 308: we should check all parent folders, not just this one

tibbi:
alright, should be good now thanks

PR #2905, opened 2023-07-22
fix: lack of import statement

### LINKED ISSUE #2904: Compile Error: compile error at "app/src/main/kotlin/com/simplemobiletools/gallery/pro/extensions/Context.kt"
The project compiles error, and the error message is as follows.

image

This is because this class lacks the import statement of `com.bumptech.glide.request.target.Target`, and it thinks that the `Target` class is a annotation class in kotlin-stdlib.

Maple-pro:
Fixes #2904


tibbi:
nice, thanks

PR #2907, opened 2023-07-24
Migrate to media3 exoplayer

### LINKED ISSUE #2750: Resuming/Pausing Video Inaccurate
I noticed that ever since I started using the app, resuming/pausing video playback seems to be inaccurate and would always "autocorrect" itself in a way that ignores the video's milliseconds duration.

I have also confirmed that this is the case for all other supported video file formats.

Here's a demonstration:

https://user-images.githubusercontent.com/32636532/214215067-3d8ed07b-7357-4237-8d19-6a5bf93beb9a.mp4

esensar:
This migrates current exoplayer2 to `androidx.media3`. Everything seems to be working the same, just with different package names.

This closes #2750, by preventing exoplayer cleanup on every pause (it should only be cleaned up when it is going to be destroyed).


tibbi:
alright thanks, thought this will be more work :D

PR #2909, opened 2023-07-24
Handle password protection when copying/moving folders

### LINKED ISSUE #2901: Protected Folders can be moved to other folder without entering pin
If a folder is protected with a pin inside the app you can simply move the folder to an unprotected folder without entering the pin.
All pictures are than accessible within the unprotected folder.

Any change to a protected folder should require the pin.

App version: 6.27.0

esensar:
This changes copy and move handling to work similarly to delete. It will ask for protection if single folder is selected and it will ignore protected folders if multiple are selected.

This closes #2901


tibbi:
nice thanks, seems to work well

PR #2915, opened 2023-07-28
Add separate options for directory grouping

### LINKED ISSUE #1886: Subfolder Grouping Issue

esensar:
Since I can see valid uses for both of these options, I believe it is best to offer both options to the users. File structure grouping will get familiar behavior which will just display directories as they are in the file system, whereas direct subfolder grouping does exactly as it says (groups only direct subfolders - if we were to change it to more than direct, we are moving in the direction of file structure).

This closes #1886

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/gallery/pro/activities/MainActivity.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/gallery/pro/activities/MainActivity.kt
@@ -299,7 +299,7 @@ class MainActivity : SimpleActivity(), DirectoryOperationsListener {
     override fun onBackPressed() {
         if (main_menu.isSearchOpen) {
             main_menu.closeSearch()
-        } else if (config.groupDirectSubfolders) {
+        } else if (config.isDirectoryGroupingActive()) {
             if (mCurrentPathPrefix.isEmpty()) {
                 super.onBackPressed()
             } else {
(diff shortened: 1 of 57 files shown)


tibbi:
omg this got way too confusing... Do not add any new settings or strings, just revert to the previous version with the bugs fixed please

esensar:
> omg this got way too confusing... Do not add any new settings or strings, just revert to the previous version with the bugs fixed please

There are no bugs to be fixes. The subfolder grouping originally worked as intended, but some people did not like it. I will just revert it, but there will still be that original issue that some people don't like that grouping as it may be confusing.

esensar:
Revert PR: https://github.com/SimpleMobileTools/Simple-Gallery/pull/2924

tibbi:
there was a bug when in some cases some folders didnt get grouped properly. It might depend on the order folders are processed or whatever, not sure

esensar:
> there was a bug when in some cases some folders didnt get grouped properly. It might depend on the order folders are processed or whatever, not sure

Do we have it reported somehwere? #1886 pretty much just describes confusion with this feature. I did not notice bugs there, but maybe I am wrong.

tibbi:
the issue describes it properly. Some folders are not grouped, while some are. Maybe the folder casing it causing it like dcim/DCIM or who knows what.

PR #2924, opened 2023-07-31
Revert "Make subfolder grouping follow file structure"

esensar:
This reverts commit 11aa25ad45f4f466251a53c86e6e1eb0402091d5.


tibbi:
Ill accept it until it gets fixed properly

esensar:
I think we need to define how exactly we expect it to behave, to be able to fix it properly.

Currently, it works as advertised - it groups direct subfolders. If there are gaps in between (folders with no media files), these get ignored and their children appear in the top view. If we go for filling the gaps, we end up in the behavior which just follows file structure directly.

PR #2925, opened 2023-07-31
Pre-fill parent folders when grouping subfolders

esensar:
This ensures that folders without media files
which have direct child folders with media don't get different treatment during grouping. This should make subfolder grouping a bit more predictable.

### DIFF
--- a/app/src/main/kotlin/com/simplemobiletools/gallery/pro/extensions/Context.kt
+++ b/app/src/main/kotlin/com/simplemobiletools/gallery/pro/extensions/Context.kt
@@ -42,6 +42,7 @@ import java.io.FileInputStream
 import java.nio.ByteBuffer
 import java.nio.channels.FileChannel
 import kotlin.collections.set
+import kotlin.math.max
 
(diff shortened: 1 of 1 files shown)


esensar:
This will still not fix stuff like grouping images that are put in folders in a structure like `img/YEAR/EVENT` as reported in #1886. The issue with handling that is finding the cut-off point for grouping (selecting the right parent directory, since in this case, none of these have direct children with media and it is a hierarchy that is a couple of folders deep).

esensar:
Try the first example from #1886 (Folder A with folders 1,2,3 and folder B with folders X,Y,Z)

tibbi:
well, that grouping already works to me, without the PR too

tibbi:
actually I see, lets hope it works well, thanks

esensar:
I think we will still have to figure something out for this. I think that grouping in formats like `year/month/day` or `year/month/event` are pretty common, and this setting will not help with that, since this only considers direct subfolders, but at least this will cover some of the issues people had without breaking existing functionality.

PR #2936, opened 2023-08-05
support animated thumbnails for animated webp files

gmasil:
Current behaviour:
* GIFs have animated thumbnails if the option "animate GIFs at thumbnails" is selected
* WEBP files have no animated thumbnail
* GIFs have no rounded corners when both options "rounded corners" and "animate GIFs at thumbnails" are selected, breaking the design
* thumbnails of static GIFs have no crossfade animation
* no way to control if folder thumbnails should be animated

Behaviour with this PR:
* GIFs have animated thumbnails if the option "animate GIFs at thumbnails" is selected
* WEBPs have animated thumbnails if the option "animate GIFs at thumbnails" is selected
* If both options "rounded corners" and "animate GIFs at thumbnails" are selected all thumbnails will have rounded corners, but will not be animated (limitation of currently used libraries) to provide an overall better look
* thumbnails of all file types are treated the same way (except for SVGs) and have the same crossfade animation
* separate options to enable animated thumbnails for folders and files


gmasil:
Sure, you can use this one: https://mathiasbynens.be/demo/animated-webp-supported.webp

gmasil:
I would like to wait until https://github.com/SimpleMobileTools/Simple-Gallery/pull/2943 is merged.
I started again based on that branch and will update this PR when 2943 is resolved.

tibbi:
merged, check conflicts please

gmasil:
I noticed that thumbnails of static GIFs are missing the crossfade animation and I saw there is different code for each file type. I aggregated all functionalities in the `loadImageBase` function that can be called for every file type (except for SVGs as they are actually different).
Additionally I added a separate option for folder thumbnail style to turn animated thumbnails on or off. So I like my folders to be static without animations, but inside the folders I like the animated thumbnails. Having those options separated allows users to select the desired behaviour more accurately.

app/src/main/kotlin/com/simplemobiletools/gallery/pro/dialogs/ChangeFileThumbnailStyleDialog.kt
```diff
@@ -59,13 +65,28 @@ class ChangeFileThumbnailStyleDialog(val activity: BaseSimpleActivity) : DialogI
...
         config.showThumbnailFileTypes = binding.dialogFileStyleShowThumbnailFileTypes.isChecked
         config.markFavoriteItems = binding.dialogFileStyleMarkFavoriteItems.isChecked
         config.thumbnailSpacing = thumbnailSpacing
     }
 
+    private fun updateAnimateGifsCheckbox() {
+        if (binding.dialogFileStyleRoundedCorners.isChecked) {
+            binding.dialogFileStyleAnimateGifs.alpha = 0.5f
+            binding.dialogFileStyleAnimateGifs.isChecked = false
+            binding.dialogFileStyleAnimateGifsHolder.setOnClickListener { Toast.makeText(activity, R.string.unsupported_animate_gifs_and_rounded_corners, Toast.LENGTH_SHORT).show() }
```
gmasil, line 83: Let me know what you think about this behaiour. If "rounded corners" is selected the option for "animate thumbnails" is grayed out and if pressed a toast is shown stating that rounded corners and animated thumbnails are not supported at the same time.
This is due to glide not supporting fit center and rounded corners at the same time, this would break the design and looks bad.

tibbi:
havent been working on this app recently, but I dont think it needs a new string if we have one for files already..

gmasil:
> havent been working on this app recently, but I dont think it needs a new string if we have one for files already..

So you mean I should remove this string?
```xml
<string name="animate_gifs_in_folders">Animate thumbnails of folders</string>
```
And just use the existing one for files and folders?

tibbi:
sure. The other string isnt needed either, just hide the options that arent available

gmasil:
Okay I removed the strings again, the options are now simply hidden when unavailable like other options do.

app/src/main/kotlin/com/simplemobiletools/gallery/pro/dialogs/ChangeFileThumbnailStyleDialog.kt
```diff
@@ -1,11 +1,14 @@
 package com.simplemobiletools.gallery.pro.dialogs
 
 import android.content.DialogInterface
+import android.widget.Toast
```
tibbi, line 4: this is no longer needed, always optimize imports and code style before the PR

gradle/libs.versions.toml
```diff
@@ -22,6 +22,7 @@ androidImageCropper = "4.5.0"
 apng = "2.25.0"
 awebp = "2.25.0"
 glideCompiler = "4.15.1"
+zjupureWebpdecoder = "2.3.4.15.1"
```
tibbi, line 25: why do we need a new library if animating gifs in files works already?

gmasil: Animating thumbnails of GIFs is working, but not for WEBPs. This library is a glide extension to extend support for WEBP files. This is required to get animated thumbnails for animated WEBP files working.

app/src/main/kotlin/com/simplemobiletools/gallery/pro/dialogs/ChangeFileThumbnailStyleDialog.kt
gmasil: Sorry, I am still getting used to IntelliJ. Did it now for all changed files.

app/src/main/kotlin/com/simplemobiletools/gallery/pro/extensions/Context.kt
```diff
@@ -37,14 +41,14 @@ import com.simplemobiletools.gallery.pro.interfaces.*
...
 import com.squareup.picasso.Picasso
-import pl.droidsonroids.gif.GifDrawable
 import java.io.File
 import java.io.FileInputStream
 import java.nio.ByteBuffer
 import java.nio.channels.FileChannel
 import kotlin.collections.set
 import kotlin.math.max
 
+
```
tibbi, line 51: avoid adding duplicate empty lines

tibbi:
yeah, it actually works well. But to avoid duplication, lets move the "Animate GIFs at thumbnails" setting from both file and folder thumbnail style to the main settings page, below "Crop thumbnails into squares" and just use it for both files and folders. Use the old config ANIMATE_GIFS for it.

gmasil:
But then you wouldn't be able to select different options for files and folders. I personally prefer to have my folders without animated thumbnails, but my files with animated thumbnails. If we have a combined option something like that wouldn't be possible anymore. I think it es better to have separate options, so everyone can decide for themselves.

tibbi:
I dont like such redundant settings, lets merge them into one.

gmasil:
It's not redundant, just let the user decide

tibbi:
I dont want to have a million different settings..

gmasil:
I reverted the options back to original

tibbi:
lets move the setting under "Crop thumbnails into squares" as it now affects both files and folders, not just files

tibbi:
ye it is good now, thanks :)

tibbi:
showing SVGs doesnt work now, can you check if maybe this broke it?

tibbi:
@gmasil have you maybe checked the SVGs already, if the issue was caused by this PR or not?

naveensingh:
Confirming that the SVG issue was not caused by this PR, it was a KSP issue: https://github.com/SimpleMobileTools/Simple-Gallery/pull/3004

gmasil:
@naveensingh thanks for checking. I missed the notifications. Is this resolved now?

naveensingh:
@gmasil yep!

PR #2943, opened 2023-08-09
Migrate to viewbinding and SDK 34

esensar:
- Replaced kotlinx sythetic with viewbinding
- Updated kotlin from 1.7.10 to 1.8.22 (imgly SDK doesn't seem to be compatible with 1.9.0 at the moment)
- Updated room to 2.6.0-alpha03
- Updated Android Gradle Plugin to 8.1.0
- Upgraded `targetSDK` to 34
- Handled new media permissions (https://developer.android.com/about/versions/14/changes/partial-photo-video-access)
    - There is now additional menu option on SDK 34+ to allow access to more media - since users can select an option not to allow all
    
## TODO
- Migrate to kotlin build scripts
- Wait for https://github.com/SimpleMobileTools/Simple-Commons/pull/1784 and update `Simple-Commons` ref
- Wait for https://github.com/SimpleMobileTools/Simple-Commons/pull/1792 and update `Simple-Commons` ref


app/src/main/kotlin/com/simplemobiletools/gallery/pro/activities/EditActivity.kt
```diff
@@ -92,19 +89,21 @@ class EditActivity : SimpleActivity(), CropImageView.OnCropImageCompleteListener
     private var oldExif: ExifInterface? = null
     private var filterInitialBitmap: Bitmap? = null
     private var originalUri: Uri? = null
+    private lateinit var binding: ActivityEditBinding
```
tibbi, line 92: lets use the same binding in all apps for consistency, check calculator

gmasil:
Just as a small feedback: I was using this version over the weekend and everything worked just fine, found no bugs or weird stuff.

app/src/main/kotlin/com/simplemobiletools/gallery/pro/activities/EditActivity.kt
```diff
@@ -92,19 +89,20 @@ class EditActivity : SimpleActivity(), CropImageView.OnCropImageCompleteListener
     private var oldExif: ExifInterface? = null
     private var filterInitialBitmap: Bitmap? = null
     private var originalUri: Uri? = null
+    private val binding by viewBinding(ActivityEditBinding::inflate)
```
tibbi, line 92: use calculators binding for consistency please

tibbi: nevermind, forgot viewBinding is a helper extension..

app/src/main/res/menu/menu_main.xml
```diff
@@ -58,6 +58,11 @@
         android:showAsAction="never"
         android:title="@string/create_new_folder"
         app:showAsAction="never" />
+    <item
+        android:id="@+id/access_more_media"
+        android:showAsAction="never"
+        android:title="Allow access to more media"
```
tibbi, line 64: what is this for? It should have a proper string, not a hardcoded one

gradle/imglysdk.gradle
```diff
@@ -0,0 +1,42 @@
+// This file had to stay in Groovy
+// Since just converting it to Kotlin will not work
+// There needs to be a way to conditinally apply this plugin
+// But that results in missing references for `imglyConfig`
```
tibbi, line 4: so how does it work now, is the library properly excluded in foss build variant?

app/src/main/res/menu/menu_main.xml
esensar: Yeah, I need to define a proper string, just need to finalize the wording.
This is related to new media permissions - users can now select only a subset images they want to allow our app access to. Based on guidelines in https://developer.android.com/about/versions/14/changes/partial-photo-video-access we are supposed to let users manually allow access to more media at any point (without constantly prompting them on startup), so this option just displays the permission popup to the user to let them add more images or allow all.

gradle/imglysdk.gradle
esensar: Yeah, because this file is included in the build only when in proprietary build (using the same condition that was used previously).

app/src/main/kotlin/com/simplemobiletools/gallery/pro/activities/MainActivity.kt
```diff
@@ -345,20 +346,22 @@ class MainActivity : SimpleActivity(), DirectoryOperationsListener {
...
 
-        main_menu.getToolbar().menu.apply {
+        binding.mainMenu.getToolbar().menu.apply {
             findItem(R.id.temporarily_show_hidden).isVisible = !config.shouldShowHidden
             findItem(R.id.stop_showing_hidden).isVisible = (!isRPlus() || isExternalStorageManager()) && config.temporarilyShowHidden
 
             findItem(R.id.temporarily_show_excluded).isVisible = !config.temporarilyShowExcluded
             findItem(R.id.stop_showing_excluded).isVisible = config.temporarilyShowExcluded
+
+            findItem(R.id.access_more_media).isVisible = isUpsideDownCakePlus()
```
tibbi, line 364: this should check some permissions too, not just the OS version. Like the button is useless for example at upgrading from older app versions as everything needed is granted already. Also if the user grants access to all media files or all files in general, this menu item shouldnt be visible.

tibbi: that file picker system doesnt make sense in the gallery actually, it was supposed to be used in other apps. And it doesnt work either... If I pick some images the app can access, it still closes and shows the Storage permission is required toast. But runs well on next try.

esensar: I have removed it now. Now the app will always display that toast unless user selects `Allow all`.

tibbi:
Doesnt look like the best UX, Id rather wait until more apps update to targetSDK 34. Maybe we can revert back to 33 to have at least viewbinding and library updates included?

esensar:
Yeah, we can do that. This change will affect the app regardless, since this change affects all apps, not just ones targeting SDK 34. It will result in us always asking for permission when the app is started until user gives access to all photos.

esensar:
Maybe better to leave SDK at 34 and just revert these changes for this popup? Since the behavior will be the same

esensar:
Actually, it is not exactly the same. For lower SDK version, even selecting just one image works fine and lets the app access all the images (although temporarily only, user will be asked again when the app is started).

tibbi:
how does an app like https://play.google.com/store/apps/details?id=gallery.photomanager.picturegalleryapp.imagegallery still use the old permissions dialog on Android 14? Do they target an even lower SDK version? Not sure how to check it.

esensar:
It seems they are targeting SDK 31, so I guess it is related to that.

tibbi:
okay thanks, the app on master branch is weird on Android 14 too, so guess this will be an improvement. Will test other apps later.

tibbi:
so is this PR ready for accepting? It is still marked as draft

esensar:
> so is this PR ready for accepting? It is still marked as draft

I forgot mark it as ready, since I was waiting for resolution on these permissions.

gradle/libs.versions.toml
```diff
@@ -0,0 +1,84 @@
...
+androidx-swiperefreshlayout = { module = "androidx.swiperefreshlayout:swiperefreshlayout", version.ref = "androidx-swiperefreshlayout" }
+androidx-constraintlayout = { module = "androidx.constraintlayout:constraintlayout", version.ref = "androidx-constraintlayout" }
+#exif
+exif = { module = "it.sephiroth.android.exif:library", version.ref = "exif" }
+#Room
+androidx-room-runtime = { module = "androidx.room:room-runtime", version.ref = "room" }
+androidx-room-ktx = { module = "androidx.room:room-ktx", version.ref = "room" }
+androidx-room-compiler = { module = "androidx.room:room-compiler", version.ref = "room" }
+#Simple Mobile Tools
+simple-tools-commons = { module = "com.github.esensar:Simple-Commons", version.ref = "simple-commons" }
```
gmasil, line 58: I think you missed this in the review

tibbi: thanks, will fix it

Issue #2146, opened 2021-05-26
Performance issue WEBP


Nabokov86:
Starting from version 6.16, the application consumes significantly more resources when displayed WEBP in full-screen mode. When the WEBP is displayed, CPU heats up to 70 degrees, then the throttling starts.

Before the version where WebP animation support was added, there is no performance issue.

App version: 6.16.0 - 6.20.1
Device: arm, android 7 and 9

tibbi:
do you have a rooted device, or one modified in a similar way?

Nabokov86:
@tibbi 
My apologies for the late reply.

Yes I have a rooted device. You need a logcat, right? For some reason, I can't get logcat on my device. If I can figure it out, I'll send it to you. Sorry for that.

tibbi:
I dont need logcat, that wouldnt help. Does it work well in other apps?

Nabokov86:
@tibbi Yes, works well in other galleries. And as I mentioned earlier, everything is fine in the older versions. Now I downgraded on version 6.15.6, work great.

tibbi:
retry with 6.26.8 once it becomes available

Nabokov86:
@tibbi Sorry for the delay! I have just installed version 6.26.8 and unfortunately I am still able to reproduce it.

gmasil:
I can reproduce this issue. Static webp images produce significant more CPU usage and even with animated webp files you can see that more fluent animations will be less heavy on CPU than more slideshow like animations.
I will investigate and try to fix this.

tibbi:
should be fixed in 6.28.x

PR #2954, opened 2023-08-24
fix bug where static webp images will cause high CPU usage

gmasil:
Regarding issue https://github.com/SimpleMobileTools/Simple-Gallery/issues/2146

The option `setLoopLimit(0)` seems to make the image viewer repeatedly show the frame as fast as possible if the webp has only one frame. Removing the option has no other impact, animated webps are repeated just as expected.


tibbi:
have you tested if this hasnt affected animated files?

gmasil:
Yes I did, animated WEBP files are still being animated and repeated.
I even tested other file formats like GIFs, they are not effected as well. Still animated and still repeated.

PR #2955, opened 2023-08-25
close inputstream and datasource to prevent leaks

gmasil:
I noticed logcat shows several logs regarding resources not being closed:
```
2023-08-25 13:34:29.860 11220-11233 System                  com.simplemobiletools.gallery.pro    W  A resource failed to call close.
```
So I added this to show stacktraces to unclosed resources to fix them:
```java
        StrictMode.setVmPolicy(VmPolicy.Builder()
            .detectLeakedClosableObjects()
            .penaltyLog()
            .build())
```
I found two occurrences that do not close resources:
* check if a picture should be treated as a panorama
* data source for videos

This PR closes those resources to prevent leaks.


tibbi:
thanks, should be good

PR #2967, opened 2023-09-06
Include simple-commons locally using composite build

naveensingh:
To include the local simple-commons project, simply uncomment the following lines in settings.gradle.kts:
```kotlin
includeBuild("../Simple-Commons") {
    dependencySubstitution {
        substitute(module("com.github.SimpleMobileTools:Simple-Commons")).using(project(":commons"))
    }
}
```
Modify the path to simple-commons project as required. This can be reduced to `includeBuild("../Simple-Commons")` only as mentioned in the TODO but more research is required (might even need to add a new TXT DNS record so we can use commons like `com.simplemobiletools:simple-commons`).

This is cooler than the previous `implementation(project(":commons"))`, it allows you to develop individual projects like simple-commons and simple-gallery [in the same window](https://docs.gradle.org/7.0/userguide/composite_builds.html#composite_build_ide_integration) without any other changes:

Including commons like this will make testing changes much easier.

**Required changes:**
- Remove duplicated view ids and use in-app dividers if they are accessed programmatically (to avoid binding issues when using commons locally, was racking my brains to fix this one)
- Update AGP to 8.1.1 (composite builds must use the same plugin version)
- Update commons

**Related links:**
 - https://docs.gradle.org/7.0/userguide/composite_builds.html#composite_build_intro
- https://docs.gradle.org/current/userguide/structuring_software_products.html#defining_custom_project_types_as_convention_plugins
- https://docs.gradle.org/current/samples/sample_structuring_software_projects.html

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -17,12 +17,14 @@ dependencyResolutionManagement {
         maven(url = "https://artifactory.img.ly/artifactory/imgly")
     }
 }
-// TODO: This will be deprecated in future. Migrate to the newer `pluginManagement { includeBuild() }` mechanism instead of explicitly substituting dependency.
-//includeBuild("../Simple-Commons") {
-//    dependencySubstitution {
-//        substitute(module("com.github.SimpleMobileTools:Simple-Commons")).using(project(":commons"))
-//    }
-//}
+
 rootProject.name = "Simple-Gallery"
 enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")
 include(":app")
+
+// TODO: This will be deprecated in future. Migrate to the newer `pluginManagement { includeBuild() }` mechanism instead of explicitly substituting dependency.
+/*includeBuild("../Simple-Commons") {
+    dependencySubstitution {
+        substitute(module("com.github.SimpleMobileTools:Simple-Commons")).using(project(":commons"))
+    }
+}*/
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -18,11 +18,11 @@ dependencyResolutionManagement {
     }
 }
 // TODO: This will be deprecated in future. Migrate to the newer `pluginManagement { includeBuild() }` mechanism instead of explicitly substituting dependency.
-includeBuild("../Simple-Commons") {
-    dependencySubstitution {
-        substitute(module("com.github.SimpleMobileTools:Simple-Commons")).using(project(":commons"))
-    }
-}
+//includeBuild("../Simple-Commons") {
+//    dependencySubstitution {
+//        substitute(module("com.github.SimpleMobileTools:Simple-Commons")).using(project(":commons"))
+//    }
+//}
 rootProject.name = "Simple-Gallery"
 enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")
 include(":app")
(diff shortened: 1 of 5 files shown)


naveensingh:
If this is accepted, then https://github.com/SimpleMobileTools/Simple-Commons/pull/1818 should be merged too

gmasil:
Hey this is awesome! I tried to get this to work and suddenly you come around with the solution!

What do you think about making this kind of "autodetect" if Simple-Commons exists?

```kotlin
if(File("../Simple-Commons").exists()) {
    includeBuild("../Simple-Commons") {
        dependencySubstitution {
            substitute(module("com.github.SimpleMobileTools:Simple-Commons")).using(project(":commons"))
        }
    }
}
```

This works well as long as no one has another project called Simple-Commons with different contents.

naveensingh:
Thanks. Yeah, I thought about that but we may not always want to use the local repo even if it exists, it maybe out of date. It's best if one enables this explicitly. @tibbi what do you think?

tibbi:
we should definitely use the jitpack one, local inclusion should be commented out

gmasil:
Yeah good point, let's leave it as is.

settings.gradle.kts
```diff
@@ -17,4 +17,12 @@ dependencyResolutionManagement {
...
     }
 }
+// TODO: This will be deprecated in future. Migrate to the newer `pluginManagement { includeBuild() }` mechanism instead of explicitly substituting dependency.
+//includeBuild("../Simple-Commons") {
+//    dependencySubstitution {
+//        substitute(module("com.github.SimpleMobileTools:Simple-Commons")).using(project(":commons"))
+//    }
+//}
+rootProject.name = "Simple-Gallery"
+enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")
```
tibbi, line 27: what are these 2 lines needed for?

naveensingh: > `rootProject.name = "Simple-Gallery"`

Gradle: "Checking out the project in different folders will impact the generated code and implicitly the buildscript classpath, breaking caching."

> `enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")`

Type safe project accessors allow you to replace `implementation(project(":feature:module"))` with `implementation(project.feature.module)`. It is enabled implicitly by `includedBuild()` so I just put it there.

tibbi:
moved it at the bottom and made easier to uncomment out. What about that TODO, we cannot use the new method yet as it would require more work? Those 2 new lines `//rootProject.name = "Simple-Gallery"` and 
`//enableFeaturePreview("TYPESAFE_PROJECT_ACCESSORS")` dont seem to be needed, it works without them too

naveensingh:
> Those 2 new lines...dont seem to be needed, it works without them too

yes, it'll work without them but not setting the project name will break gradle caching for included projects as mentioned above, it's just a build performance thing.

> What about that TODO, we cannot use the new method yet as it would require more work?

Yes, it'll require some more work as the packages are supposed to have the same metadata (group ID, etc.) so gradle knows what dependency is replaced by which included project.

If you can add a DNS TXT record that maps `git.simplemobiletools.com` to `https://github.com/simplemobiletools`, I can try to make it work. Adding a record doesn't take much time but I guess you've already done that before. I already tried this but my domain name doesn't match my github username so it didn't work. More info: https://jitpack.io/docs/#custom-domain-name

**The method we are using is not deprecated yet so I don't see any problem if we keep using it.**

tibbi:
ok, so Ill accept it as it is, thanks

naveensingh:
okay :)

PR #2997, opened 2023-10-05
Fixed setting group ordering (#2996)

### LINKED ISSUE #2996: Impossible to choose "descending" order in Group By
Selecting descending order in the Group By dialog does nothing. It doesn't respect the setting and when, going back to the dialog, Ascending is selected instead.

I am using 6.28.0 from the Play Store. I also tested 6.26 from f-droid and it works correctly.

Aga-C:
Fixes #2996


tibbi:
nice, thanks

